PR #1, opened 2021-08-03
Robustify against negative dt values

Timple:


include/path_tracking_pid/path_tracking_pid_local_planner.hpp
```diff
@@ -29,6 +29,7 @@
 #include <visualization_msgs/MarkerArray.h>
 
 #define MAP_PARALLEL_THRESH 0.2
+constexpr double DT_MAX=1.5;
```
cesar-lopez-mar, line 32: this is in seconds right?

Timple: Yup, always SI

cesar-lopez-mar (approved):

PR #11, opened 2022-01-20
Fix/harvey 939 acc dec switched in reverse

MCFurry:


src/controller.cpp
```diff
@@ -586,7 +586,8 @@ geometry_msgs::Twist Controller::update(const double target_x_vel,
   // to de-accelerate and thus avoid abrupt velocity changes at the end of the trajectory
   // The sample time plays an important role on how good these estimates are.
   // Thus We add a distance to the end phase distance estimation depending on the sample time
-  if (current_x_vel > target_end_x_vel)
+  if ((sign(current_target_x_vel_) > 0 && current_x_vel > target_end_x_vel) ||
+      (sign(current_target_x_vel_) < 0 && current_x_vel < target_end_x_vel))
```
Rayman, line 590: Should this be `|| (sign(current_target_x_vel_) < 0 && current_x_vel < -target_end_x_vel))`

Maybe it's better to do this with absolute values

Rayman: And also, comparisons like `sign(...) < 0` is equivalent to `... < 0`

MCFurry: > Should this be `|| (sign(current_target_x_vel_) < 0 && current_x_vel < -target_end_x_vel))`

No, we check the case where we have negative velocity and our current velocity is 'too big'. `target_end_x_vel` is signed, so we can have a plan to arrive with `target_end_x_vel=-1.0` and if we drive -2.0 now, we should 'decelerate'.

src/controller.cpp
```diff
@@ -630,18 +631,42 @@ geometry_msgs::Twist Controller::update(const double target_x_vel,
...
+    {
+      current_target_acc = -target_x_decc_;
+    }
+    else
+    {
+      current_target_acc = target_x_decc_;
+    }
   }
-  else if (current_x_vel > current_target_x_vel_)
+  else if (sign(current_target_x_vel_) > 0)         // Positive velocity requested
```
Rayman, line 647: `sign` is here not needed

test/test_path_tracking_pid_fw_bw.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Start goal and evaluate outcome
+        outcome_exp = rospy.get_param("~outcome", GS.SUCCEEDED)
+        self.reconfigure(2.0)
+        rospy.logwarn("Starting path!")
+        client.send_goal(ExePathGoal(path=path))
+
+        rospy.sleep(0.5)
+        self.assertTrue(3.5 < self.cur_accel < 4.5, msg="Violated acceleration {}".format(self.cur_accel))
+        rospy.sleep(4.5)
```
Rayman, line 59: I'm a bit worried about these sleeps. Are you using `use_sim_time` during testing? If not these sleeps can take longer due if there is more load on the runner

MCFurry: Nope we're not, but we're using `mobile_robot_simulator` here which simply runs on ROS-clock (opposed to i.e. gazebo who publishes the ROS-clock).

Timple: We do have issues with these things @Rayman. Therefor the margin should be bigger than desired, in case CI is slow.

But since this is in line with the other tests I'm good here.

Timple (approved):

PR #15, opened 2022-02-01
fix: Use target x velocity to determine collision look ahead direction

PaulVerhoeckx:
**Issue:**
When stopping for an obstacle whilst driving backwards, the robot will come to a standstill. Thereafter the configured `collision_look_ahead_length_offset` is applied forwards as  `copysign(0) = 1` (see [here](https://github.com/nobleo/path_tracking_pid/blob/main/src/path_tracking_pid_local_planner.cpp#L424)). This results in the robot creeping towards the obstacle until the footprint reaches the obstacle (see gif)

**Solution:**
Use the target velocity (`target_x_vel`) instead.    

path_tracking_pid collision

### DIFF
--- a/src/path_tracking_pid_local_planner.cpp
+++ b/src/path_tracking_pid_local_planner.cpp
@@ -421,7 +421,8 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
 
   // Define a x_step transform which will be used to step forward the position.
   tf2::Transform x_step_tf;
-  x_step_tf.setOrigin(tf2::Vector3(copysign(x_resolution, x_vel), 0.0, 0.0));
+  double target_x_vel = pid_controller_.getConfig().target_x_vel;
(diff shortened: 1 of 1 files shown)


cesar-lopez-mar (approved):
Looks good to me

PaulVerhoeckx:
I have added a commit where I use the absolute maximum value of the actual and target velocity. For example on a cancel request `target_x_vel` is set to `0`, whilst the actual velocity is still not `0`.

Timple:
So in hindsight, couldn't we have used the sign of the target velocity? Instead of the maximum of both.

That way we have a single source of velocity.

Or we can simply keep the sign correct.

PaulVerhoeckx:
That's indeed how I implemented in the first commit not sure why I changed it. Possibly because `target_x_vel` is set to zero on a preempt, applying `collision_look_ahead_length_offset` forwards. But in that case we're stopping as fast as possible anyway.

PR #20, opened 2022-02-02
fix: check predicted footprints for LETHAL_OBSTACLE instead of INSCRIBED_INFLATED_OBSTACLE

### LINKED ISSUE #16: Controller is too conservative in its collision checking
The controller checks if there is any cell within the collision polygon that is of value `costmap2d::INSCRIBED_INFLATED_OBASTACLE`. If there is, it concludes that a collision is imminent. This is not true. the robot footprint is perfectly allowed to enter space of `costmap2d::INSCRIBED_INFLATED_OBASTACLE` value. Here's an example where there is overlap, but no collision:

no_collision

The cheap way to check for imminent collisions, assuming a (approximately) circular base is to check if the cell in which (a predicted) `base_link` is located is `>= costmap2d::INSCRIBED_INFLATED_OBASTACLE`. That would mean a guaranteed collision. 

However, with a non-circular base, this is not enough. Because a different orientation may also lead to collision before `base_link` reaches this space. Instead, you can check if any of the cells in the collision polygon exceeds this value. That means it reaches a value of `>= costmap2d::LETHAL_OBSTACLE`.

rokusottervanger:
Fixes #16


Timple:
Hmm, bummer. The test fails.

This is because the collision that is expected never happens. The slow-down behavior is such that maximum-slowdown is already achieved in the light-blue area (INSCRIBED_INFLATED_OBSTACLE). So the robot keeps crawling about until a timeout occurs.

The nice part of the inflation layer was that is had a nice gradient to it. Now there is an inflation present with a fixed value (INSCRIBED_INFLATED_OBSTACLE). 

Not sure how this gradient can be re-introduced here... :/

rokusottervanger:
The same test still fails, but now because of the tracking error. We now check the costmap value at the location of `base_link` and the projected predictions of `base_link`. Because of this, the velocity remains higher for a longer time. This results in a tracking error in the U turn. This happens in every U-turn test, but this test just happens to stop before the controller converges.

IMO, this is a drawback of the tests, and not a problem this PR should solve. I think we should split up this test script into more atomic test cases. This test's responsibility is not to check the tracking error, but if stopping occurs when collision is imminent.

rokusottervanger:
Fixed the tests anyway ;)

Timple:
> We now check the costmap value at the location of base_link and the projected predictions of base_link

So on a long rectangular vehicle, do we slow down more if the center is passing an obstacle, than when a corner is passing the obstacle?

rokusottervanger:
Yes. I'm afraid that's how costmap2d is designed. In such a case, the inflation radius can be tuned to be high enough for the emerging behavior to be acceptable.

Timple:
We do have vehicles which a significantly longer than wide. So having them slow down just a little when corners are close to obstacles seems like quite a regression :(

Can we fix this so that the obstacle <-> vehicle distance is still used, instead of the obstacle <-> base_link?

rokusottervanger:
I discussed with @Rayman, and we found that this is a shortcoming of the assumptions made in the design of costmap2d. 

In the ideal situation, the costmap would contain only distances to obstacles, providing the user (a controller or planner) with the freedom to determine its own cost related to proximity. Implementing this would render our controller incompatible with normal use of the costmap. The controller would no longer be interchangeable with others.

I agree that slowing down just a little when corners are close to obstacles does not seem appropriate, but slowing down enough can still be achieved by increasing the inflation radius and/or the cost scaling factor (the exponent of the cost decay function). The question then becomes: would this slowdown be acceptable (performance-wise) also when it is not strictly necessary because the robot happens to have a more favorable orientation?

cesar-lopez-mar:
I do recognize this shortcoming of costmap2d. Especially in applications where there are objects all around, like warehouses, and we use a very asymmetric robot (IDA, JAX are also examples). I do think the change in this PR makes more sense than the current implementation for the reasons given by @rokusottervanger. But I think we need to change the logic how we decelerate when the robot is moving towards obstacles. An idea is that instead of using the costmap values to compute the new velocities, we should use the current velocity, a range of desired decelerations and  distance to obstacles along the direction of movement. In any case we can open a new issue and design together such a strategy.

rokusottervanger:
I agree. And for now, to get the desired speed reduction, we can just increase the `inflation_radius` and decrease the `cost_scaling_factor` with this PR merged in.

cesar-lopez-mar (approved):

rokusottervanger:
@Timple I wouldn't want to merge without your approval. If there's anything else you'd need, let me know, so I can make changes accordingly.

Timple:
Thanks for waiting!

> And for now, to get the desired speed reduction, we can just increase the inflation_radius and decrease the cost_scaling_factor with this PR merged in.

That won't give the exact same behavior as before right?

Timple:
Thanks for the explanation. Lets await tomorrows discussion :+1:

Timple (approved):

Timple (approved):

PR #23, opened 2022-02-10
Refactor visualization to a separate class

Rayman:
Right now the visualization code is mixed with math. This leads to a large file. This PR is the first step into moving visualization code into a separate class.

I've changed the marker type slightly (to spheres) as I find it easier to see the error.
Screenshot from 2022-02-10 10-59-43

### DIFF
--- a/include/path_tracking_pid/visualization.hpp
+++ b/include/path_tracking_pid/visualization.hpp
@@ -1,0 +1,30 @@
+#include <string>
+
+#include "geometry_msgs/Pose.h"
+#include "ros/publisher.h"
+#include "std_msgs/ColorRGBA.h"
+#include "tf2/LinearMath/Transform.h"
+
+namespace path_tracking_pid
+{
+class Visualization
+{
+public:
+  Visualization(ros::NodeHandle nh);
+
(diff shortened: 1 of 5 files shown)


include/path_tracking_pid/visualization.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+namespace path_tracking_pid
+{
+class Visualization
+{
+public:
+  Visualization(ros::NodeHandle nh);
+
+  void publishControlPoint(std_msgs::Header header, tf2::Transform pose);
+  void publishAxlePoint(std_msgs::Header header, tf2::Transform pose);
```
Timple, line 16: Curious, why not send a TransformStamped?

Timple (approved):
Thanks!

include/path_tracking_pid/visualization.hpp
Rayman: I think most of the internal datatypes should be `tf2::Transform` because you can perform calculations with them. I could also change this to `tf2::Stamped<tf2::Transform>` if you want

Timple: Well, bonus of you current implementation is that the header only needs initialisation once in PID.

PR #25, opened 2022-02-15
Fixed signed/unsigned comparisons.

lewie-donckers:
Fixed signed/unsigned comparisons.

This allows us to make the compiler warning/error settings stricter.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -104,7 +104,7 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   // For now do not allow repeated points or in-place rotation
   // To allow that the way the progress is checked and the interpolation is done needs to be changed
   // Also check if points suddenly go in the opposite direction, this could lead to deadlocks
-  for (int pose_idx = 1; pose_idx < global_plan.size() - 1; ++pose_idx)
+  for (std::vector<geometry_msgs::PoseStamped>::size_type pose_idx = 1; pose_idx < global_plan.size() - 1; ++pose_idx)
   {
     auto prev_pose = global_plan[pose_idx - 1].pose;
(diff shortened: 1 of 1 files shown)


src/controller.cpp
```diff
@@ -104,7 +104,7 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   // For now do not allow repeated points or in-place rotation
   // To allow that the way the progress is checked and the interpolation is done needs to be changed
   // Also check if points suddenly go in the opposite direction, this could lead to deadlocks
-  for (int pose_idx = 1; pose_idx < global_plan.size() - 1; ++pose_idx)
+  for (std::vector<geometry_msgs::PoseStamped>::size_type pose_idx = 1; pose_idx < global_plan.size() - 1; ++pose_idx)
```
Timple, line 107: This is really explicit. Wouldn't simply `size_t` suffice?

Timple (approved):

src/controller.cpp
lewie-donckers: Probably. But this is not guaranteed. See https://en.cppreference.com/w/cpp/container/vector

A vector's `size_type` could be something other than  `(std::)size_t`.

As far as I'm concerned this is temporary as we don't want any raw for loops. We should prefer proper algorithms or alternatively range based for loops.

Timple: True

Rayman: I think this is quite dangerous. If `pose_idx` is an unsigned int, that means that this comparison can underflow: `pose_idx < global_plan.size() - 1`

Rayman: Google style guide recommends using ints everywhere, except when doing bit manipulations

PR #26, opened 2022-02-15
Stricter compiler settings.

lewie-donckers:
Stricter compiler settings.

Will be made even stricter when I have more changes in.

PS: This is a stacked PR. All will be merged to main one by one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.0.2)
+add_compile_options(-std=c++17 -Wall -Wextra -Wno-unused-parameter -Werror)
 project(path_tracking_pid)
 
 ## Find catkin and any catkin packages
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -67,6 +66,7 @@ add_library(${PROJECT_NAME}
   src/visualization.cpp
 )
 add_dependencies(${PROJECT_NAME} ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS} ${PROJECT_NAME}_gencfg)
+target_compile_options(${PROJECT_NAME} PUBLIC -std=c++17 -Wall -Wextra -Wno-unused-parameter -Werror)
```
Timple, line 69: Is there a benefit to do this on a per-target basis? Instead of the complete scope of the package like it was before?

lewie-donckers: As far as I know, this is a CMake best practice. Other libs/packages might need different compiler settings. Especially now we're making the compiler stricter. Other nobleo libs/packages might not build with that.

Timple: But other libs/packages are not influenced by the `add_compile_options` in this CMakeLists.txt. It's confined to this package only.

We do have packages with lots of nodes, I think having the same settings within a package makes the most sense. Otherwise it becomes a tedious effort to align the settings even withing a package.

Reason for asking is that `add_compile_options` is the recommended ROS practice (I'll try and find the source for this later).

lewie-donckers: Fixed.

Timple (approved):

PR #28, opened 2022-02-15
Replaced macros with constants.

lewie-donckers:
Replaced all macros with constants.

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -19,9 +19,9 @@ template <typename T> int sign(T val) {
 namespace path_tracking_pid
 {
 
-#define RADIUS_EPS 0.001  // Smallest relevant radius [m]
-#define VELOCITY_EPS 1e-3  // Neglegible velocity
-#define LONG_DURATION 31556926  // A year (ros::Duration cannot be inf)
+inline constexpr double RADIUS_EPS = 0.001; // Smallest relevant radius [m]
+inline constexpr double VELOCITY_EPS = 1e-3; // Neglegible velocity
+inline constexpr double LONG_DURATION = 31556926; // A year (ros::Duration cannot be inf)
 
 enum class ControllerMode
 {
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -19,9 +19,9 @@ template <typename T> int sign(T val) {
 namespace path_tracking_pid
 {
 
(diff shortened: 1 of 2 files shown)


Timple:
I'm surprised this passes the build.

Note to self: check why roslint is not active, there should be two spaces before the comment.

rokusottervanger:
It's an improvement. But more important may be to check why these are constants and not parameters in the first place.

lewie-donckers:
> It's an improvement. But more important may be to check why these are constants and not parameters in the first place.

I suggest you have someone with more domain knowledge than me create an issue about this.

rokusottervanger:
> > It's an improvement. But more important may be to check why these are constants and not parameters in the first place.
> 
> I suggest you have someone with more domain knowledge than me create an issue about this.

You're right. This discussion is already going on in #2 . I'll create an issue about magic numbers in general.

Timple:
> Is this something you want fixed?

I want the build to fail on this :).

But yes, please fix as well.

include/path_tracking_pid/controller.hpp
```diff
@@ -19,9 +19,9 @@ template <typename T> int sign(T val) {
 namespace path_tracking_pid
 {
 
-#define RADIUS_EPS 0.001  // Smallest relevant radius [m]
-#define VELOCITY_EPS 1e-3  // Neglegible velocity
-#define LONG_DURATION 31556926  // A year (ros::Duration cannot be inf)
+inline constexpr double RADIUS_EPS = 0.001; // Smallest relevant radius [m]
+inline constexpr double VELOCITY_EPS = 1e-3; // Neglegible velocity
+inline constexpr double LONG_DURATION = 31556926; // A year (ros::Duration cannot be inf)
```
Rayman, line 24: Does it need to be marked inline? Or is this needed because this constant is used in multiple compilation units?

Rayman (commented):
Nice, constants are definitely much better than macros

lewie-donckers:
> > Is this something you want fixed?
> 
> I want the build to fail on this :).
> 
> But yes, please fix as well.

Fixed by formatting those lines according to the .clang-format file provided by @Timple.

include/path_tracking_pid/controller.hpp
lewie-donckers: Yep. That's what it is for. If you define a static member variable or a variable at namespace scope in the header file, you should make it inline.

Rayman (approved):

PR #29, opened 2022-02-15
Moved constants to reduce scope.

lewie-donckers:
Reduced the scope of constants.

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es5-keep-scopes-small

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -4,6 +4,8 @@
 
 #include "path_tracking_pid/controller.hpp"
 
+#include "common.hpp"
+
 #include <limits>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


src/controller.cpp
```diff
@@ -13,6 +15,13 @@
 namespace path_tracking_pid
 {
 
+namespace {
+
+constexpr double RADIUS_EPS = 0.001; // Smallest relevant radius [m]
+constexpr double LONG_DURATION = 31556926; // A year (ros::Duration cannot be inf)
+
+} // namespace anonymous
```
Timple, line 23: What's the benefit of this construction?

lewie-donckers: Unnamed (anonymous) namespaces in source files ensure any declarations/definitions in that namespace are only available to that source file. It is a superior version of declaring stuff static in source files.

See:
- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities
- https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces
- https://stackoverflow.com/questions/357404/why-are-unnamed-namespaces-used-and-what-are-their-benefits

rokusottervanger (approved):

Timple (approved):

PR #33, opened 2022-02-15
Clarify `init_vel_max_diff` check

Timple:


README.md
```diff
@@ -102,7 +102,7 @@ Tracking_pid parameters are all available through (rqt_)dynamic_reconfigure. The
 * **`l`** (double, default: `0.5`) Following distance from robot's rotational point to trajectory.
 * **`track_base_link`** (bool, default: `false`) Whether to track the path using the base_link instead of the control point ahead. A smooth path is needed.
 * **`init_vel_method`** (enum, default: `1`) Method to choose initial velocity on new paths.
-* **`init_vel_max_diff`** (double, default: `0.5`) How much velocity difference is acceptable upon starting a new path. If internal state and new path's velocity differ more than this, the path will be rejected. Set to -1 to ignore this check.
+* **`init_vel_max_diff`** (double, default: `0.5`) How much velocity difference is acceptable upon starting a new path. If internal state and new path's velocity differ more than this, the path will be rejected. Set to -1 to ignore this check. This check is only active if 'init_vel_method'==InternalSetpoint.
```
rokusottervanger, line 105: The path does not have a velocity.

"If the last velocity setpoint and the current actual velocity differ more than this..."

This should be updated if we decide to move this check to the `computeVelocityCommands` implementation, as the goal will then not be rejected, but aborted.

rokusottervanger (changes requested):

README.md
Timple: > This should be updated if we decide to move this check to the computeVelocityCommands implementation, as the goal will then not be rejected, but aborted.

I think this is already done now

rokusottervanger: Still, if you're updating this line, might as well fix this as well, right?
```suggestion
* **`init_vel_max_diff`** (double, default: `0.5`) How much velocity difference is acceptable upon starting a new path. If internal state and current odometry's velocity differ more than this, the path will be aborted. Set to -1 to ignore this check. This check is only active if 'init_vel_method'==InternalSetpoint.
```

rokusottervanger (approved):

PR #36, opened 2022-02-15
Distinguish between system and normal headers.

lewie-donckers:
Distinguish between system header files (which are less strictly checked by the compiler and tools like clang-tidy) and normal header files. System header files should be used for header files which are not part of the current project/repo.

This allows us to use properly strict compiler settings.


Timple:
Note to self: See if `cfgbuild.cmake` is required

Timple (approved):

PR #37, opened 2022-02-15
Even stricter compiler settings.

lewie-donckers:
Properly strict compiler settings.

And required fixes.


Rayman (approved):

src/path_tracking_pid_local_planner.cpp
```diff
@@ -520,9 +520,9 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
   return max_cost;
 }
 
-uint32_t TrackingPidLocalPlanner::computeVelocityCommands(const geometry_msgs::PoseStamped& pose,
-                                                          const geometry_msgs::TwistStamped& velocity,
-                                                          geometry_msgs::TwistStamped& cmd_vel, std::string& message)
+uint32_t TrackingPidLocalPlanner::computeVelocityCommands(const geometry_msgs::PoseStamped& /* pose */,
```
MCFurry, line 523: What is the idea of making these arguments `unnamed` ?

Rayman: Strict compilers warn for unused arguments. This is a method to still name them, and you can see that they are unused.

rokusottervanger (approved):

src/path_tracking_pid_local_planner.cpp
MCFurry: Ahh I see, they just didn't all seem 'unused' to me

Timple: Well, if that's a surprise we might need another look 👍 
But anyhow, this PR can continue I think.

Timple (approved):

src/path_tracking_pid_local_planner.cpp
lewie-donckers: @MCFurry please resolve if you agree.

PR #40, opened 2022-02-15
Cleaned up includes.

lewie-donckers:
Cleaned up includes according to the following guidelines:

- use <> for files found through include dirs and "" for files found relative to the current file see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else
- regarding the order:
  - first include the corresponding header file (this one is for source files only)
  - then local files (using "")
  - then files from the same lib
  - then files from 3rd party libs (boost/ros/etc)
  - then standard includes
  - then OS includes

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -1,17 +1,17 @@
 #pragma once
 
+#include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/Transform.h>
+#include <geometry_msgs/Twist.h>
+#include <path_tracking_pid/PidConfig.h>
+#include <path_tracking_pid/PidDebug.h>
+#include <tf2/LinearMath/Transform.h>
+
 #include <array>
+#include <boost/noncopyable.hpp>
 #include <path_tracking_pid/details/fifo_array.hpp>
 #include <path_tracking_pid/details/second_order_lowpass.hpp>
 #include <vector>
-
-#include "boost/noncopyable.hpp"
-#include "geometry_msgs/PoseStamped.h"
-#include "geometry_msgs/Transform.h"
-#include "geometry_msgs/Twist.h"
-#include "path_tracking_pid/PidConfig.h"
-#include "path_tracking_pid/PidDebug.h"
-#include "tf2/LinearMath/Transform.h"
 
 namespace path_tracking_pid
 {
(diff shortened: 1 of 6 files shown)


Rayman (commented):
One month ago I did this PR which is in almost the opposite direction: https://github.com/nobleo/path_tracking_pid/pull/14.

There I've used a different header style (`<` vs `"`). So the question is which style do we use?

I've followed the ROS2 code style which refers to the google code style. It's not really clear but for me it boils down to:

```c++
#include "current_package/header.hpp"

#include <system_headers>

#include "other_ros_headers/header.hpp"
#include "current_package/other_header.hpp"
```
What do you think?

lewie-donckers:
> One month ago I did this PR which is in almost the opposite direction: #14.
> 
> There I've used a different header style (`<` vs `"`). So the question is which style do we use?
> 
> I've followed the ROS2 code style which refers to the google code style. It's not really clear but for me it boils down to:
> 
> ```c++
> #include "current_package/header.hpp"
> 
> #include <system_headers>
> 
> #include "other_ros_headers/header.hpp"
> #include "current_package/other_header.hpp"
> ```
> 
> What do you think?

@Rayman the order I suggested is mostly my personal preference. The most important thing about the order is to include the corresponding header file first in every source file. Other than that - assuming the header files are correct - the order should not matter. So I'll happily use whatever guideline you guys want.

About the brackets to use ("" and <>), I recommend we do follow the standard convention which is also in the C++ Core Guidelines.

Please let me know what order you guys want to standardize in our ROS projects. @Rayman @Timple @rokusottervanger.

Timple:
ROS1 linter doesn't care about `"` or `<`.

The ROS2 clang formatter uses `<`. So I propose to go for these?

When I'm on my work-pc I can try `ament_clang_format --reformat` and verify it's output.
_But frankly I don't mind much 🙂_

Timple:
So I noticed the ROS2 formatter doesn't care about `<` and `"`. 

> use <> for files found through include dirs and "" for files found relative to the current file

This does seem to be the main difference between the two styles. So I propose to merge this in line with that.

@Rayman any last thoughts?

lewie-donckers:
I just tried the ROS2 clang-format file provided by @Timple : https://github.com/ament/ament_lint/blob/master/ament_clang_format/ament_clang_format/configuration/.clang-format

And it will automatically group and sort all includes. (Using an odd order but at least it's standardized.)

Assuming we want to use that clang-format file, we just have to agree on the "" and <> brackets.

Timple (approved):

PR #41, opened 2022-02-15
Explicitly disabled copy/move.

lewie-donckers:
Explicitly disabled copy/move for `Controller` and `TrackingPidLocalPlanner`.

This to avoid accidental copies which could be incorrect/expensive/both.

And to avoid accidental moves which could cause bound callbacks to break.


include/path_tracking_pid/details/no_copy_no_move.hpp
```diff
@@ -0,0 +1,20 @@
...
+{
+public:
+  NoCopyNoMove(const NoCopyNoMove&) = delete;
+  NoCopyNoMove(NoCopyNoMove&&) = delete;
+  NoCopyNoMove& operator=(const NoCopyNoMove&) = delete;
+  NoCopyNoMove& operator=(NoCopyNoMove&&) = delete;
+
+protected:
+  NoCopyNoMove() = default;
+  ~NoCopyNoMove() = default;
```
Rayman, line 17: Why does a `NoCopyNoMove` class need a default constructor?

lewie-donckers: It needs a default constructor because it is used as a base class, so it needs to be instantiated (without any arguments).

We need to explicitly default it because we declared (deleted) the copy and move constructors which inhibits the compiler from generating a default constructor itself.

Timple:
> [](https://github.com/lewie-donckers)This to avoid accidental copies which could be incorrect/expensive/both.
And to avoid accidental moves which could cause bound callbacks to break.

This sounds like a generic c++ issue for these kind of constructions.
Isn't there a standard library providing this?

lewie-donckers:
> > This to avoid accidental copies which could be incorrect/expensive/both.
> > And to avoid accidental moves which could cause bound callbacks to break.
> 
> This sounds like a generic c++ issue for these kind of constructions. Isn't there a standard library providing this?

@Timple No, not in the standard library. Boost does have a similar implementation. Since it's so easy to implement and I wasn't sure about the Boost policy, I went for my own.

Want to change to Boost?

Timple:
Preferably. Boost is used a lot by ROS anyway, so I don't mind the extra include.

This construction (and the logic) is a bit hard to follow. And I'm affraid it's also tricky to reproduce (and knowing when) in other packages.

Reducing the code in this PR will increase the chance of that.

include/path_tracking_pid/details/no_copy_no_move.hpp
lewie-donckers: The compiler will generate a default constructor but only under certain conditions. And because we declared the copy c'tor and move c'tor (with ` = delete;`), it is not allowed to do so in this case.

I tried to explain that with:

> We need to explicitly default it because we declared (deleted) the copy and move constructors which inhibits the compiler from generating a default constructor itself.

Rayman: In my reply I talked about the *destructor*

lewie-donckers: Oopsie. Indeed the destructor is always available. But we want it to be protected, so we have to default it explicitly.

Base classes should always have a (public) virtual destructor or a protected non-virtual destructor.

Rayman: OK, clear!

Rayman:
Which boost utility should we use then? I could only find the following:
```
BOOST_MOVABLE_BUT_NOT_COPYABLE
BOOST_COPYABLE_AND_MOVABLE
boost::noncopyable
```

lewie-donckers:
> Preferably. Boost is used a lot by ROS anyway, so I don't mind the extra include.
> 
> This construction (and the logic) is a bit hard to follow. And I'm affraid it's also tricky to reproduce (and knowing when) in other packages.
> 
> Reducing the code in this PR will increase the chance of that.

@Timple fixed!

Rayman (approved):

Timple (approved):
Thanks!

I do think this is just as readable in the controller source code. But it reduces code maintenance and increases the chance of it being re-used.

Timple:
> And to avoid accidental moves which could cause bound callbacks to break.

This could be added as an in-code comment for future reference. It's an important reason to use this construction I think.

Issue #35, opened 2022-02-15
Get rid of magic numbers


rokusottervanger:
In a few places, there's magic numbers in the code. We may want to get rid of those. Some examples:

https://github.com/nobleo/path_tracking_pid/blob/14958134bf07aa30256f83c399c481fee41f22b6/include/path_tracking_pid/controller.hpp#L22-L24

https://github.com/nobleo/path_tracking_pid/blob/14958134bf07aa30256f83c399c481fee41f22b6/include/path_tracking_pid/path_tracking_pid_local_planner.hpp#L24-L25

https://github.com/nobleo/path_tracking_pid/blob/14958134bf07aa30256f83c399c481fee41f22b6/include/path_tracking_pid/controller.hpp#L330-L338

Timple:
> #define RADIUS_EPS 0.001  // Smallest relevant radius [m] 
>  #define VELOCITY_EPS 1e-3  // Neglegible velocity 

These are meant for numerical comparisons. Like in c++ you cannot compare floats directly. But good practice is not to use EPS itself but choose a domain-specific smallest number. So I think these can stay without being parameterised.

>  #define LONG_DURATION 31556926  // A year (ros::Duration cannot be inf)

This is a workaround for INF missing in the message. For Robot domain I think we're good here.


I agree that all others can be parameterised. 

---

Is it worth doing though? Without someone actually asking for this, it just makes the package more complex.

lewie-donckers:
Just to be 100% sure we're talking about the same thing:

A magic value is an unnamed literal value which is directly used in the code. The examples you listed above are constants and that's fine. Maybe they can be replaced with variables, removed or given more descriptive names but they are not magic values.

Not a magic value:
```cpp
double upper_limit_ = 100.0; 
```

Magic value:
```cpp
      if (abs(new_nominal_x_vel) < 0.01)
```

rokusottervanger:
I stand corrected. Nonetheless, at least _some_ of these values should be made configurable for the user. Do you have a suggestion for an alternative issue title?

PR #44, opened 2022-02-15
Replaced member variables with constants.

lewie-donckers:
Replaced member variables that were never modified with constants.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -20,6 +20,21 @@ namespace {
 constexpr double RADIUS_EPS = 0.001;        // Smallest relevant radius [m]
 constexpr double LONG_DURATION = 31556926;  // A year (ros::Duration cannot be inf)
 
+// Cutoff frequency for the derivative calculation in Hz.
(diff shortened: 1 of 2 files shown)


src/controller.cpp
```diff
@@ -21,6 +21,21 @@ namespace {
...
+constexpr double lat_upper_limit = 100.0;
+constexpr double lat_lower_limit = -100.0;
+
+constexpr double ang_upper_limit = 100.0;
+constexpr double ang_lower_limit = -100.0;
+
+// Anti-windup term. Limits the absolute value of the integral term.
+constexpr double windup_limit = 1000.0;
+
+
```
Rayman, line 38: @cesar-lopez-mar are these parameters the same for all robots, or should they be user configurable?

rokusottervanger: I created a more generic issue for this question in #35

Timple (approved):

rokusottervanger (approved):

PR #45, opened 2022-02-15
Ignore Python cache files.

lewie-donckers:
Ignore python cache files. These are created when running the tests.


Rayman (approved):

Rayman:
Yes, but this is an open source package, so users might not have a nobleo gitignore

Timple:
> Yes, but this is an open source package, so users might not have a nobleo gitignore

I agree, they might not have.
But then they also have this issue on all of their ROS packages.

Aside from that I'm good for pycache, as this is package-specific.

Can we move the file to the `test` folder? That would make it clear where these files are expected and why we're ignoring them.

lewie-donckers:
> > Yes, but this is an open source package, so users might not have a nobleo gitignore
> 
> I agree, they might not have. But then they also have this issue on all of their ROS packages.
> 
> Aside from that I'm good for pycache, as this is package-specific.
> 
> Can we move the file to the `test` folder? That would make it clear where these files are expected and why we're ignoring them.

@Timple moved .gitignore to tests.

Timple (approved):

PR #47, opened 2022-02-15
Quaternion of carrotTF was not initialized, set to proper value

MCFurry:
As initiated here:
https://github.com/nobleo/path_tracking_pid/commit/f603fecff527321a9ee0165845981bd4fc578ad6#commitcomment-66762408
Building with proper flags threw a new error, this fixes it.

A `tf2::Transform` without arguments is not initialized, when calling `setOrigin()` the rotation part is still not initialized.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -132,8 +132,8 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   if (!track_base_link_enabled_)
   {
     // Add carrot length to plan using goal pose (we assume the last pose contains correct angle)
-    tf2::Transform carrotTF(tf2::Matrix3x3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0));
-    carrotTF.setOrigin(tf2::Vector3(l_, 0.0, 0.0));
+    tf2::Transform carrotTF(tf2::Matrix3x3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0),
+                            tf2::Vector3(l_, 0.0, 0.0));
     global_plan_tf_.push_back(last_transform * carrotTF);
   }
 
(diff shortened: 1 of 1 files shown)


Timple:
So this is surprising. I think ROS promises to initalize all messages.

And `tf2::Transform` is a ROS message right?

Although the `w=1` still needs to be done of course.

src/controller.cpp
```diff
@@ -132,7 +132,7 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   if (!track_base_link_enabled_)
   {
     // Add carrot length to plan using goal pose (we assume the last pose contains correct angle)
-    tf2::Transform carrotTF;
+    tf2::Transform carrotTF(tf2::Matrix3x3(1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0));
     carrotTF.setOrigin(tf2::Vector3(l_, 0.0, 0.0));
```
Timple, line 136: This line can be combined with the initialisation above right?

MCFurry: Nope it's a convenience class, not a message

lewie-donckers:
I'm not sure where to put this, so I'll just do this here.

I propose/recommend we start building debug and release versions of the software and test them both. This to catch any differences in compiler warnings, optimizations, initialization, asserts, etc.

@Timple @rokusottervanger

MCFurry:
> I'm not sure where to put this, so I'll just do this here.
> 
> I propose/recommend we start building debug and release versions of the software and test them both. This to catch any differences in compiler warnings, optimizations, initialization, asserts, etc.
> 
> @Timple @rokusottervanger

Good thinking! I think we have to extend the pipeline to do that?

lewie-donckers (approved):

src/controller.cpp
MCFurry: > This line can be combined with the initialisation above right?

Yup that's even neater of course, will do!

Timple (approved):

Timple:
@lewie-donckers Moved your discussion to the board. You should have an invitation.

PR #50, opened 2022-02-16
Removed dead code.

lewie-donckers:
Removed dead code.

`cutoff_frequency_lat` and `cutoff_frequency_ang` are constants and both have the value -1. So the code that was only run if they were unequal to -1 could be removed. This in turn effectively made the member variables `c_lat_ ` and `c_ang_` constants.


rokusottervanger:
Lgtm, but I think @cesar-lopez-mar should comment as he implemented the filter and may want this in. In that case, it should be revived by making it user-configurable. If it has never been necessary before, we should indeed get rid of it.

Timple:
@cesar-lopez-mar You can add an approval in the 'Files changed' tab.

cesar-lopez-mar (approved):

PR #51, opened 2022-02-16
Remove dead code.

lewie-donckers:
Removed `Controller::selectMode()` and the related enum because they were not used anywhere.


Timple:
@cesar-lopez-mar Is this right that they are unused?

cesar-lopez-mar:
Correct, the user determines which loops to use in the parameters

cesar-lopez-mar (approved):

Issue #48, opened 2022-02-16
Possible unsigned underflow


Rayman:
PR #25 changed this line:

https://github.com/nobleo/path_tracking_pid/blob/b5c139119c74602a2421fec024e3cded41fac821/src/controller.cpp#L122

@lewie-donckers I think this is quite dangerous. If `pose_idx` is an unsigned int, that means that this comparison can underflow when global plan is empty.

_Originally posted by @Rayman in https://github.com/nobleo/path_tracking_pid/pull/25#r806908279_

Timple:
Lets fix this. Or assert it to be non-zero if it cannot be.

PR #52, opened 2022-02-16
Fixed signed/unsigned conversion warnings.

lewie-donckers:
Fixed signed/unsigned conversion warning from clang-tidy.

`warning: narrowing conversion from 'unsigned long' to signed type 'int' is implementation-defined [bugprone-narrowing-conversions]`

Ideally the loops should later be refactored further into algorithm calls or range based for loops.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -185,38 +185,36 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   tf2::Transform deltaPlan;
   // We define segment0 to be the segment connecting pose0 and pose1.
   // Hence, when picking the starting path's pose, we mean to start at the segment connecting that and the next pose.
-  for (int idx_path = global_plan_tf_.size() - 2; idx_path >= 0; --idx_path)
+  for (auto idx_path = global_plan_tf_.size() - 1; idx_path > 0; --idx_path)
   {
     /* Get distance to segment to determine if this is the segment to start at */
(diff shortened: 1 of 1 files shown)


src/controller.cpp
```diff
@@ -185,38 +185,36 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
   tf2::Transform deltaPlan;
   // We define segment0 to be the segment connecting pose0 and pose1.
   // Hence, when picking the starting path's pose, we mean to start at the segment connecting that and the next pose.
-  for (int idx_path = global_plan_tf_.size() - 2; idx_path >= 0; --idx_path)
+  for (auto idx_path = global_plan_tf_.size() - 1; idx_path > 0; --idx_path)
```
Rayman, line 188: Did you see https://github.com/nobleo/path_tracking_pid/issues/48? I'm worried about unsigned underflow. Especially with loops going backward

lewie-donckers: Yes, you're right.

I accidentally changed the behavior of the code:
- it was incorrect if plan size was very large (>= INT_MAX + 3)
- it now is incorrect if the plan size is 0.
- 
The latter is more likely, so I'll change this back to an int and add a static cast to silence the warning.

lewie-donckers: Fixed.

Rayman (approved):

Rayman:
Fixes #48

Rayman (approved):

PR #53, opened 2022-02-16
Ensure all members are initialized.

lewie-donckers:
Ensure all members are initialized.

Since the members were previously uninitialized and the software seemed to work fine, the actual value of the initialized variables apparently is not important. So i just went with 0.

clang-tidy can check this. I will add a clang-tidy file once its warning have been fixed.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -227,13 +227,13 @@ private:
   void printParameters();
 
   path_tracking_pid::PidConfig local_config_;
-  ControllerState controller_state_ = ControllerState();
+  ControllerState controller_state_;
 
   // Global Plan variables
   std::vector<tf2::Transform> global_plan_tf_;  // Global plan vector
   std::vector<double> distance_to_goal_vector_;  // Vector with distances to goal
   std::vector<double> turning_radius_inv_vector_;  // Vector with computed turning radius inverse
-  double distance_to_goal_;
+  double distance_to_goal_ = 0;
   tf2::Transform current_goal_;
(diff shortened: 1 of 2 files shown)


Rayman:
Question: all these values are probably initialized, but not immediately where they are declared. Just a little bit later. Can't `clang-tidy` see further into the code to notice nothing is wrong?

lewie-donckers:
> Question: all these values are probably initialized, but not immediately where they are declared. Just a little bit later. Can't `clang-tidy` see further into the code to notice nothing is wrong?

It warns about the members not being initialized in the constructor. Which is - to say the least - a best practice. Since the objects are not constructed very often, there are no noticeable performance penalties.

Timple:
I'm not worried about performance. But a developer might now assume they are used.
Perhaps using `-1` and `NaN` as default will make clear that later overwrites are intended?

lewie-donckers:
> I'm not worried about performance. But a developer might now assume they are used. Perhaps using `-1` and `NaN` as default will make clear that later overwrites are intended?

That's fine with me. But I will need some help then. Which members should get which value? (Feel free to commit required changes.)

Timple:
All `int` a `-1` and all doubles and floats NaN.
Unfortunately, int and bool don't have NaN :)

Rayman:
If sonarcloud can check these constructs, I would leave the code as is and remove the `clang-tidy` check. But I'm fine with changing code as well.

lewie-donckers:
> All `int` a `-1` and all doubles and floats NaN. Unfortunately, int and bool don't have NaN :)

Fixed!

Timple (approved):

PR #55, opened 2022-02-16
Added clang-tidy and fixed all warnings.

lewie-donckers:
Added a clang tidy configuration file. Made a lot of minor changes to fix all clang tidy warnings.

All relevant clang tidy warning sets are enabled but some of the warnings within those sets have been disabled. Eventually, all of them should be enabled, I think. But for now, this is a very good start.

I also propose to:
- add similar settings to the other projects
- run clang-tidy in the (automated) build
- provide documentation and/or scripts so developers can run it themselves

But I think that should be done in other PRs.

If  you want to run this yourself, make sure it is installed (`apt install clang-tidy`) and run:
```
clang-tidy -p ../build/path_tracking_pid/ `find include src -type f`
```

The output should look like this:
```
rosdev@76e35e075b2d:/workspaces/nobleo-ros-dev_path_tracking_pid/src$ clang-tidy -p ../build/path_tracking_pid/ `find include src -type f`
25664 warnings generated.
79883 warnings generated.
111570 warnings generated.
143725 warnings generated.
144313 warnings generated.
175459 warnings generated.
232007 warnings generated.
Suppressed 232007 warnings (232005 in non-user code, 2 with check filters).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
```

This means it detected a lot of warnings but those were all in the system headers. If it really finds something, it will show an error or warning similar to clang/gcc compiler errors/warnings.


include/path_tracking_pid/controller.hpp
```diff
@@ -98,11 +98,10 @@ class Controller : private boost::noncopyable
    * @return tf of found position on plan
    * @return index of current path-pose if requested
    */
-  tf2::Transform findPositionOnPlan(const geometry_msgs::Transform current_tf,
-                                    ControllerState* controller_state_ptr,
-                                    size_t &path_pose_idx);
+  tf2::Transform findPositionOnPlan(geometry_msgs::Transform current_tf, ControllerState* controller_state_ptr,
+                                    size_t& path_pose_idx);
```
Rayman, line 102: Shouldn't most of ROS message arguments be const references?
```c++
const geometry_msgs::Transform& current_tf
```

lewie-donckers: Yeah, I think so. I checked `geometry_msgs::Transform` and it is quite large. Too large to pass by value, I'd say.

clang-tidy has a warning for inefficient argument passing and it fired on another function, so I assumed it would have checked everything. Apparently it is unable to do this for every function or it uses different heuristics. I will check all functions tomorrow.

lewie-donckers: @Rayman fixed!

src/controller.cpp
```diff
@@ -258,14 +258,14 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
 
 // https://stackoverflow.com/questions/849211/shortest-distance-between-a-point-and-a-line-segment
 // TODO(Cesar): expand to 3 dimensions
-double Controller::distSquared(const tf2::Transform& pose_v, const tf2::Transform& pose_w) const
+double Controller::distSquared(const tf2::Transform& pose_v, const tf2::Transform& pose_w)
```
Rayman, line 261: Just curious, why isn't this const anymore?

Rayman (approved):

Timple (approved):

PR #57, opened 2022-02-16
Added FifoArray abstraction.

lewie-donckers:
I noticed there was a lot of code duplication to keep track of error history in `ControllerState`. I have 3 PRs lined up to address this. This one is the first.

Added a FIFO array abstraction.

Please note (and check) the changed offsets for the filtered errors. The old code already had moved all entries in that buffer when it calculated the new value. The new code does not so it needs different offsets.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <path_tracking_pid/details/fifo_array.hpp>
 
 #include <array>
 #include <vector>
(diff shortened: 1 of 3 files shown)


Timple:
Aligning all these similar constructs is a good idea.

Same thing here, can this complete header file be replaced with an include from the boost circular buffer?
https://www.boost.org/doc/libs/1_55_0/doc/html/circular_buffer/example.html

lewie-donckers:
> Aligning all these similar constructs is a good idea.
> 
> Same thing here, can this complete header file be replaced with an include from the boost circular buffer? https://www.boost.org/doc/libs/1_55_0/doc/html/circular_buffer/example.html

No, not easily. We need a fixed-size buffer (not just fixed-capacity) like `std::array` because of the way it is used. Wrapping the boost circular buffer in order to make it seem fixed-size would introduce more noise, I think.

Timple (approved):

PR #58, opened 2022-02-16
Added filtered error tracker abstraction.

lewie-donckers:
I noticed there was a lot of code duplication to keep track of error history in ControllerState. I have 3 PRs lined up to address this. This one is the second.

Added a filtered error tracker abstraction. The errors in `ControllerState` are always tracked in pairs. The "raw" errors and the filtered errors. The filtered errors are always calculated in the same way and only need the raw errors and previous filtered errors as input. This abstracts that principle to reduce code duplication and noise in the `Controller`.

### DIFF
--- a/include/path_tracking_pid/details/filtered_error_tracker.hpp
+++ b/include/path_tracking_pid/details/filtered_error_tracker.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+#include <path_tracking_pid/details/fifo_array.hpp>
+
+namespace path_tracking_pid::details
+{
+
+// Error tracker for the last 3 error and filtered error values.
+class FilteredErrorTracker
+{
+public:
+  // Pushes the given value to the errors FIFO buffer. A corresponding filtered error value is calculated and pushed
+  // to the filtered errors FIFO buffer.
+  void push(double value);
+
+  // Resets both errors and filtered errors FIFO buffers.
+  void reset();
+
+  // Read-only access to the errors FIFO buffer.
+  const FifoArray<double, 3>& errors() const;
+
+  // Read-only access to the filtered errors FIFO buffer.
+  const FifoArray<double, 3>& filtered_errors() const;
+
+private:
+  FifoArray<double, 3> errors_;
+  FifoArray<double, 3> filtered_errors_;
+};
+
+}  // namespace path_tracking_pid::details
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -30,11 +30,6 @@ constexpr double ang_lower_limit = -100.0;
 // Anti-windup term. Limits the absolute value of the integral term.
 constexpr double windup_limit = 1000.0;
 
-// Used in filter calculations. Default 1.0 corresponds to a cutoff frequency at
-// 1/4 of the sample rate.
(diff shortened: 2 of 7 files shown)


Timple:
I'm surprised we use both the raw errors and the filtered ones.

In my opinion we need only the filtered. Post processing of sensor data is very common and typically used to reduce sensor noise.
So the filtered one is the best estimate for the real-world value. And thus the only interesting one.

Adding @cesar-lopez-mar as he might have an opinion about this.

rokusottervanger:
Looking at the calculations you need at least a buffer of raw error measurements to calculate the new filtered one. That sounds correct if you want to use higher derivatives in your filtering.

Timple:
Ow yes, they are needed at some point internally. But not outside of the class.

So actually I'm proposing to remove the read-only access to the raw errors.

src/controller.cpp
```diff
@@ -509,55 +504,30 @@ geometry_msgs::Twist Controller::update(double target_x_vel,
 
   // trackin_error here represents the error between tracked link and position on plan
   controller_state_.tracking_error_lat = current_tracking_err.y();
-  controller_state_.tracking_error_ang = controller_state_.error_ang.at<0>();
+  controller_state_.tracking_error_ang = controller_state_.error_ang.errors().at<0>();
```
cesar-lopez-mar, line 507: Here the raw errors are used, instead of the filtered ones. We should first filter the error, and then use that one in the P I D actions. This can be done in a separate PR

cesar-lopez-mar:
I see we do use the raw errors to compute the integral action, we should align the code to use only the filtered errors in the P,I,D components. This can be done in a separate PR

src/controller.cpp
rokusottervanger: Would you make an issue for that then, and approve this PR?

rokusottervanger (approved):

Rayman:
What kind of filter is this? For a low-pass you don't need a buffer of measurements, only the previous filtered value and the next raw value.

cesar-lopez-mar (approved):

cesar-lopez-mar:
> What kind of filter is this? For a low-pass you don't need a buffer of measurements, only the previous filtered value and the next raw value.

It is a second order filter. The filter was originally implemented in this repo: https://github.com/nobleo/tracking_pid

In the comments they made a reference to "Julius O. Smith III, Intro. to Digital Filters With Audio Applications"

include/path_tracking_pid/details/filtered_error_tracker.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <path_tracking_pid/details/fifo_array.hpp>
+
+namespace path_tracking_pid::details
+{
+
+// Error tracker for the last 3 error and filtered error values.
+class FilteredErrorTracker
```
Rayman, line 9: Can we rename this class then to `SecondOrderLowpass`. This makes it really clear what's happening here

lewie-donckers: @Rayman fixed!

Rayman (approved):

PR #59, opened 2022-02-16
Added derivative filtered error tracker abstraction.

lewie-donckers:
I noticed there was a lot of code duplication to keep track of error history in ControllerState. I have 3 PRs lined up to address this. This one is the third.

Added a derivative filtered error tracker abstraction. The remaining errors in `ControllerState` are also always tracked in pairs. The 
raw errors/filtered errors and their derivatives. The derivatives are always calculated in the same way. This abstracts that principle to reduce code duplication and noise in the `Controller`.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,12 @@
+Checks: 'boost-*,
+         bugprone-*,
+         cert-*,-cert-err58-cpp,
+         clang-*,-clang-analyzer-core.uninitialized.Assign,
+         concurrency-*,
+         cppcoreguidelines-*,-cppcoreguidelines-pro-type-vararg,-cppcoreguidelines-pro-bounds-array-to-pointer-decay,-cppcoreguidelines-avoid-magic-numbers,
+         misc-*,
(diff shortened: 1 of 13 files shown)


Timple:
I think this refactor looks good code-wise. But could be much better conceptually.

Basically we only want to calculate a derivative because we're doing PID control.

Abstracting the PID controller completely makes much more sense than abstracting P, I and D seperate. Would also make the code much more readable.

lewie-donckers:
> I think this refactor looks good code-wise. But could be much better conceptually.
> 
> Basically we only want to calculate a derivative because we're doing PID control.
> 
> Abstracting the PID controller completely makes much more sense than abstracting P, I and D seperate. Would also make the code much more readable.

Sounds good but I'm lacking domain knowledge to do this myself. Do you want to finish these PRs for the small improvement they bring? Or cancel/close them and do the PID control abstraction instead (by someone else)?

Timple:
I would go for the latter.

It will probably rework most of the stuff here. The one implementing this can still use this as a small example.

Timple:
Yup

PR #60, opened 2022-02-16
Const locals when possible.

lewie-donckers:
Made local variables const if possible.

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction


Timple (approved):

Rayman (commented):
In my own side I never make local variables const because the scope is so small. The Google code style says `Using const on local variables is neither encouraged nor discouraged.`

So I'm fine with the change if you prefer this style

PR #61, opened 2022-02-16
Replaced helper function sign().

lewie-donckers:
Replaced helper function `sign()`.

I saw three (minor) problems with it:
- it has a bit of an odd (and undocumented) return type/value.
- it is basically a inferior version of std::signbit().
- it was only used in one location to check if two values have the same sign.

So I replaced it with a function that does exactly that with a descriptive name and an intuitive return type/value.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -31,10 +31,11 @@ constexpr double ang_lower_limit = -100.0;
 constexpr double windup_limit = 1000.0;
 
 
-// Typesafe sign implementation with signum:
-// https://stackoverflow.com/a/4609795
-template <typename T> int sign(T val) {
-    return (T(0) < val) - (val < T(0));
+// Indicates if both values have the same sign.
+template <typename T>
+bool have_same_sign(T val1, T val2)
+{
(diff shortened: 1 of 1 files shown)


Timple (approved):

Rayman (changes requested):
I wouldn't just remove the old `sign` function. There are a lot of places where it could be used.

If you grep for `sign` you see a lot of these constructions
```c++
copysign(1.0, somevalue)
```
Which could be replaced by
```c++
sign(somevalue)
```

lewie-donckers:
> I wouldn't just remove the old `sign` function. There are a lot of places where it could be used.
> 
> If you grep for `sign` you see a lot of these constructions
> 
> ```c++
> copysign(1.0, somevalue)
> ```
> 
> Which could be replaced by
> 
> ```c++
> sign(somevalue)
> ```

While I think refactoring those `copysign()` calls is a great idea, I don't think we should keep the function because we might need it in the future.

We know it existed (we can refer to it in an issue to request the refactor) so we can retrieve it when we want to use it again. But for now, it is dead code.

PR #62, opened 2022-02-17
Added to_transform() helper.

lewie-donckers:
Added `to_transform()` helper function to work around the output parameter of `tf2::fromMsg()`.

IMO, shows how much easier functions with return values are to use over functions with output parameters.

Will also allow me to refactor one of the manual for loops later.


Rayman (approved):
There is a reason why ROS hasn't made `fromMsg` just return the result. It's because function overloads don't work for return values and there are many possible conversions.
```
geometry_msgs::Pose <--> tf2::Transform
geometry_msgs::Transform <--> tf2::Transform
```
So making our own function for this is fine I think

Timple:
Actually, isn't is specified here already?

http://docs.ros.org/en/melodic/api/tf2/html/namespacetf2.html#a66edbda6fa9127178de05ca185603b18

Rayman:
> Actually, isn't is specified here already?
> 
> http://docs.ros.org/en/melodic/api/tf2/html/namespacetf2.html#a66edbda6fa9127178de05ca185603b18

I think that template doesn't have implementations for `geometry_msgs::Pose` because of the reason I listed

Timple:
Aah, check!

Timple (approved):

rokusottervanger (approved):

PR #63, opened 2022-02-17
Moved distSquared() helpers.

lewie-donckers:
Moved the `distSquared()` helper functions from static members to the anonymous namespace in the source file.

Will allow me to use them in other helper files in that namespace in the future.

Also refactored the implementation to avoid the duplicate sub-expressions and simply use a library function.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -49,8 +49,7 @@ tf2::Transform to_transform(const geometry_msgs::Pose & pose)
 // Returns the square distance between two points
 double distSquared(const tf2::Transform & a, const tf2::Transform & b)
 {
-  return std::pow(a.getOrigin().x() - b.getOrigin().x(), 2) +
-         std::pow(a.getOrigin().y() - b.getOrigin().y(), 2);
+  return a.getOrigin().distance2(b.getOrigin());
 }
 
 // Return the square distance between two points.
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -46,6 +46,20 @@ tf2::Transform to_transform(const geometry_msgs::Pose & pose)
   return result;
 }
 
+// Returns the square distance between two points
+double distSquared(const tf2::Transform & a, const tf2::Transform & b)
+{
+  return std::pow(a.getOrigin().x() - b.getOrigin().x(), 2) +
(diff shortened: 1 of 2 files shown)


src/controller.cpp
```diff
@@ -46,6 +46,20 @@ tf2::Transform to_transform(const geometry_msgs::Pose & pose)
   return result;
 }
 
+// Returns the square distance between two points
+double distSquared(const tf2::Transform & a, const tf2::Transform & b)
+{
+  return std::pow(a.getOrigin().x() - b.getOrigin().x(), 2) +
+         std::pow(a.getOrigin().y() - b.getOrigin().y(), 2);
```
Rayman, line 53: `tf2::Vector` has a function for this called `distance2`

src/controller.cpp
```diff
@@ -46,6 +46,20 @@ tf2::Transform to_transform(const geometry_msgs::Pose & pose)
...
+}
+
+// Return the square distance between two points.
+double distSquared(const geometry_msgs::Pose & a, const geometry_msgs::Pose & b)
+{
+  return std::pow(a.position.x - b.position.x, 2) +
+         std::pow(a.position.y - b.position.y, 2);
```
Rayman, line 60: I think our internal representation should use `tf2` datatypes because you can perform calculations with them.

Rayman: Btw, `pow` is very very heavy in unoptimized code. A self multiplication would be much faster.

Rayman: Point taken, moving these helper functions improves code quality.

Rayman (approved):

rokusottervanger (approved):

src/controller.cpp
lewie-donckers: I wanted to switch to this function but it seems that `distance2()` also takes the `z()` coordinate into account. Wouldn't that change the behavior of the function?

Rayman: When the path is received, the z-component is set to 0, so in practice this shouldn't be a difference: https://github.com/nobleo/path_tracking_pid/blob/816262318d90ac4d4da17a5ee013bc8bb6da80a2/src/path_tracking_pid_local_planner.cpp#L125

lewie-donckers: Fixed!

PR #65, opened 2022-02-17
Replaced for loop with algorithm.

lewie-donckers:
This PR replaces a raw for loop with an algorithm. In principle, you should not use raw for loops. Preferably replace them with an algorithm or at least a range-based for loop. This all to express intent: indicate what should be done rather than just how it should be done.

Sometimes (like in this case), the standard (or Boost) has no suitable algorithm. I then prefer to write my own.

I realize this change might be controversial but at the very least I wanted to show you how this can be done.

Note the following:
- instead of a raw for loop we now have a named algorithm call `transform_if()` which has its own documentation. If you're familiar with `std::transform()` or the std algorithms in general, the meaning should already be clear.
- we now use const iterators so we know the algorithm can't change anything in the source container
- it's a lot easier to check if we stay within the bounds of the source container
- both the if and transform predicates are lambdas which mostly forward to named functions, so it is clear what they do as well
- this particular change resulted in more lines of code because the algorithm had to be added (and documented). once you have your own library of algorithms (in addition to the standard ones) this will typically not be the case.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -63,6 +63,31 @@ bool is_pose_angle_obtuse(
   return distSquared(prev, next) > (distSquared(prev, cur) + distSquared(cur, next));
 }
 
+// Three-way conditional version of std::transform(). Iterates over three ranges: one defined by
+// [first1, last1), the other beginning at first2 and the last beginning at first3. For each triple
+// of elements from those ranges it applies the if-predicate and if (and only if) that returns
+// true, applies the transform-predicate and stores the result in another range, keeping the
+// original elements order and beginning at d_first.
+template <
+  typename input_iter_type, typename output_iter_type, typename if_predicate_type,
+  typename transform_predicate_type>
+output_iter_type transform_if(
+  input_iter_type first1, input_iter_type last1, input_iter_type first2, input_iter_type first3,
+  output_iter_type d_first, if_predicate_type if_predicate,
+  transform_predicate_type transform_predicate)
+{
+  while (first1 != last1) {
+    if (if_predicate(*first1, *first2, *first3)) {
+      *d_first = transform_predicate(*first1, *first2, *first3);
+      ++d_first;
+    }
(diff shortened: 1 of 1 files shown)


src/controller.cpp
```diff
@@ -156,27 +181,34 @@ void Controller::setPlan(geometry_msgs::Transform current_tf, geometry_msgs::Twi
...
+    const auto if_predicate = [](const auto & prev, const auto & cur, const auto & next) {
+      return is_pose_angle_obtuse(prev.pose, cur.pose, next.pose);
+    };
+    const auto transform_predicate = [](
+                                       const auto & /* prev */, const auto & cur,
+                                       const auto & /* next */) { return to_transform(cur.pose); };
+
+    transform_if(
+      it_begin, it_end, it_begin + 1, it_begin + 2, std::back_inserter(global_plan_tf_),
+      if_predicate, transform_predicate);
```
Rayman, line 200: I can understand the motivation for this change, however I would do it slightly different.

In the future we want all internal datatypes to be `tf2` so that we can immediately can perform calculations with them. This will make the `to_transform` call not needed.

As an intermediate step we could split this algorithm in to
```
std::transform(...)  // convert path into tf2 datatypes
std::copy_if(...is_pose_angle_obtuse(...)) // copy only obtuse angles
```
Later we could then remove the `std::transform` call.

What do you think about this method?

rokusottervanger (commented):
I understand the rationale (I don't like the index based for loop in the original), but I'm not sold on this one. That's because of two reasons. 

Firstly, the behavior changes. The user does not get feedback on which pose in the path is wrong anymore. 

Secondly, and more importantly, I'm not a c++ expert, but I consider myself a decent programmer, and I find this very hard to read. Especially more novice programmers will have a hard time understanding this.

lewie-donckers:
Based on feedback, I'll close this PR to focus on other changes.

PR #66, opened 2022-02-17
Removed config duplicate members.

lewie-donckers:
A lot of the member variables in `Controller` were copies of those in `Controller::local_config_` of type `PidConfig`.

Changes:
- Renamed `local_config_` to `config_`.
- Removed all those duplicate members and switched to their counterparts in `config_`.
- The only tricky part was `Controller::configure()` where some members had conditions/extra operations.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -216,7 +216,7 @@ private:
    */
   void printParameters() const;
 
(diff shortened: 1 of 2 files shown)


Timple (approved):
Some lines become slightly less readable because of the `config.` prepended to pretty much every variable.

But overall it does make it clear in the code that these can change at any given time. So I'm good.

rokusottervanger (approved):

PR #67, opened 2022-02-17
Used ROS2 clang-format on all files.

lewie-donckers:
Used the (official?) ROS2 clang-format file provided by @Timple on all files:

https://github.com/ament/ament_lint/blob/master/ament_clang_format/ament_clang_format/configuration/.clang-format

So this PR consists only of formatting changes.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -1,9 +1,8 @@
 #pragma once
 
+#include <array>
 #include <path_tracking_pid/details/fifo_array.hpp>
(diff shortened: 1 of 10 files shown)


Rayman (approved):
Very nice, but won't this interfere with your other PRs? :)

lewie-donckers:
> Very nice, but won't this interfere with your other PRs? :)

It's a PR train. Or stacked PRs. Whatever you want to call it. They all refer to the previous PR and continue where that one left of. If you merge them in that order, no rework is needed. If one of them is blocked (like the include files PR was) then I have to rebase the rest to remove it from the train.

Timple (approved):

PR #70, opened 2022-02-17
Preliminary custom units implementation.

lewie-donckers:
Not a proper PR so no thorough review required (for now).

I just wanted to show you what the code would look like when using a custom units library. It currently supports Speed, Acceleration, Duration, Distance and Area.

Not all members/parameters/calculations have been changed yet. Please focus on `Controller` as most changes have been made there. 

The calculation of `d_end_phase` on line 591 is particularly interesting. There a distance is calculated based on the current speed, target acceleration and target speed.

Also nice work on the original implementation because all units I have seen so far were correctly used.

I'll work on an example using the Boost units library next week. So you can compare the two.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -11,6 +11,7 @@
 #include <boost/noncopyable.hpp>
 #include <path_tracking_pid/details/fifo_array.hpp>
 #include <path_tracking_pid/details/second_order_lowpass.hpp>
+#include <path_tracking_pid/units.hpp>
 #include <vector>
 
 namespace path_tracking_pid
@@ -27,7 +28,7 @@ struct ControllerState
 {
   size_t current_global_plan_index = 0;
   size_t last_visited_pose_index = 0;
-  double current_x_vel = 0.0;
+  units::Speed current_x_vel{0};
   double current_yaw_vel = 0.0;
   double previous_steering_angle = 0.0;
   double previous_steering_x_vel = 0.0;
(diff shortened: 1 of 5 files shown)


include/path_tracking_pid/controller.hpp
```diff
@@ -177,28 +179,28 @@ class Controller : private boost::noncopyable
...
+  units::Area distToSegmentSquared(
+    const tf2::Transform & pose_p, const tf2::Transform & pose_v, const tf2::Transform & pose_w)
   {
     tf2::Transform dummy_tf;
-    double dummy_double = 0;
-    double result = 0;
-    distToSegmentSquared(pose_p, pose_v, pose_w, dummy_tf, result, dummy_double);
+    auto dummy_dist = units::Distance{0};
+    auto result = units::Area{0};
+    distToSegmentSquared(pose_p, pose_v, pose_w, dummy_tf, result, dummy_dist);
```
rokusottervanger, line 203: In this case, the result is a bit odd. We're comparing squared distances only to avoid the `sqrt` on every one of them. Technically, that results in something with unit meter squared, but it does not represent any physical area.

lewie-donckers: Yeah, I was in doubt as to what to call this. Since meters squared is the unit for area, I just went with that.

Can easily be renamed to `DistanceSquared` of course.

rokusottervanger: Downside of that, of course, is that if you __do__ want to represent an actual area, this is still represented by a `DistanceSquared`. Otherwise calculations like `distance_1 * distance_2` become ambiguous.

Alternatively, you may want to introduce a type `Length`. Multiplying two `Length` type variables would become an `Area`, and multiplying two `Distance` type variables would become a `DistanceSquared`. Similar to the distinction between `Time` and `Duration`. However, in this case, it might become somewhat ambiguous to the user. Just brainstorming here...

PR #72, opened 2022-02-21
Preliminary boost units implementation.

lewie-donckers:
Preliminary boost units implementation. No proper review required yet. This is just to show what it could look like.

It contains the same changes to the interfaces and calculations as #70  but this time using the Boost units library.

This is currently quite verbose but by adding a few typedefs we could reduce that.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -9,6 +9,7 @@
 
 #include <array>
 #include <boost/noncopyable.hpp>
+#include <boost/units/systems/si.hpp>
 #include <path_tracking_pid/details/fifo_array.hpp>
 #include <path_tracking_pid/details/second_order_lowpass.hpp>
 #include <vector>
@@ -27,7 +28,8 @@ struct ControllerState
 {
   size_t current_global_plan_index = 0;
   size_t last_visited_pose_index = 0;
-  double current_x_vel = 0.0;
+  boost::units::quantity<boost::units::si::velocity> current_x_vel =
+    0.0 * boost::units::si::meter_per_second;
   double current_yaw_vel = 0.0;
   double previous_steering_angle = 0.0;
   double previous_steering_x_vel = 0.0;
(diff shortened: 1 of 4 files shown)


lewie-donckers:
I noticed that this does not build on `melodic`. That compiler does not support the Boost units as `constexpr` values it seems.

No doubt, we could work around this, if required.

Timple:
Lets not spend any effort on melodic until we decide to do so.

src/controller.cpp
```diff
@@ -221,12 +242,12 @@ void Controller::setPlan(
   controller_state_.current_global_plan_index = 0;
 
   // find closest current position to global plan
-  double minimum_distance_to_path = 1e3;
-  double dist_to_segment = 0;
-  double iterative_dist_to_goal = 0.0;
+  auto minimum_distance_to_path = 1e3 * boost::units::si::square_meter;
```
Timple, line 245: Can't this be declared as: `boost::units::si::square_meter minimum_distance_to_path = 1e3;` ?

include/path_tracking_pid/controller.hpp
```diff
@@ -213,16 +220,17 @@ class Controller : private boost::noncopyable
...
   std::vector<double> turning_radius_inv_vector_;  // Vector with computed turning radius inverse
-  double distance_to_goal_ = 0;
+  boost::units::quantity<boost::units::si::length> distance_to_goal_{0};
   tf2::Transform current_goal_;
   tf2::Transform current_pos_on_plan_;
   tf2::Transform current_with_carrot_;
 
   // Auxiliary variables
-  double current_target_x_vel_ = 0.0;
+  boost::units::quantity<boost::units::si::velocity> current_target_x_vel_{0};
```
Timple, line 233: I think these kind of constructs can be much more readable with a typedef in the header for the used units in this package.
```
using velocity_t = boost::units::quantity<boost::units::si::velocity>

...

velocity_t current_target_x_vel_{0};
```

lewie-donckers: I completely agree. Can easily be done if/when we decide to use Boost units.

src/controller.cpp
lewie-donckers: No, the reasoning is that you could use the incorrect unit by mistake. For example write a value in centimeters, inches, feet, or whatever. So Boost units requires you to explicitly state what unit of measurement to use for the literal.

You could of course introduce a helper function to reduce verbosity:

```cpp
square_meter_t to_square_meter(double value)
{
  return value * boost::units::si::square_meter;
}
```

Timple: Aah, check. 

Amount * unit. Does make sense.

No need for the helper. This will be more readable with the typedefs I think.

rokusottervanger:
> Lets not spend any effort on melodic until we decide to do so.

We currently still have a customer using this controller under `melodic`. We're trying our best to get to move away from the system they're using that limits them from upgrading to noetic, but currently we still need to support this.

That's not to say that we need to fix this build right away, of course. Only if we decide to apply Boost units. I assume that's what you meant, @Timple ?

Timple:
Yup, if we decide to use boost units, we should make the tradeoff between the two efforts: 
Upgrading all melodic robots vs supporting melodic.

Although the latter will probably 'win'.

Timple:
Well that was an easy fix. `const` it is then!

Personally I like the boost implementation better than a self-written one. Less prone for errors in the implementation, more chance of re-use in other packages.

So as far as I'm concerned we finish this implementation in favor of #70

lewie-donckers:
I will close both PRs but keep the branches around for a bit longer.

PR #74, opened 2022-02-21
Added unittests for FifoArray.

lewie-donckers:
Added some unittests for `FifoArray<>` to get the ball rolling.

If I understand the catkin build environment correctly, these unittests will be built and run when the other tests are run. For example when you use:
```sh
catkin run_tests
```` 
or
```sh
rostest path_tracking_pid test_path_tracking_pid.test
```

You can also build them separately:
```sh
catkin build -i --make-args unittests
```

And run them separately:
```sh
rosrun path_tracking_pid unittests
```


CMakeLists.txt
```diff
@@ -106,4 +106,6 @@ install(
 
 if(CATKIN_ENABLE_TESTING)
   add_rostest(test/test_path_tracking_pid.test ARGS rviz:=false reconfigure:=false)
+  catkin_add_gtest(unittests unittest/test_main.cpp unittest/test_fifo_array.cpp)
+  target_link_libraries(unittests ${catkin_LIBRARIES})
```
Rayman, line 110: I would keep all tests in the `test` folder to keep to the standard package structure. So for example `test/unittests/...`

Rayman (changes requested):
Nice, looks good

Rayman:
- `catkin run_tests --this --no-deps` builds & runs the tests
- `catkin_test_results` gets the results of previously ran tests
- `catkin test  --this --no-deps` runs tests and gets the results

CMakeLists.txt
lewie-donckers: @Rayman fixed!

Timple (approved):
Awesome, you're getting the hang of the ROS compilation quirks/features :slightly_smiling_face:.

unittests where on the todo list for a long while now, thanks!

Rayman (approved):

PR #76, opened 2022-02-21
Refactored mandatory ptr parameter to reference.

lewie-donckers:
Refactored a mandatory (nullptr is not allowed) pointer parameter to a reference.

See:
- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f17-for-in-out-parameters-pass-by-reference-to-non-const
- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f60-prefer-t-over-t-when-no-argument-is-a-valid-option

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -94,19 +94,19 @@ public:
   /**
    * Find position on plan by looking at the surroundings of last known pose.
    * @param current Where is the robot now?
-   * @param controller_state_ptr  The current state of the controller that gets updated by this function
+   * @param controller_state  The current state of the controller that gets updated by this function
    * @return tf of found position on plan
    * @return index of current path-pose if requested
    */
   tf2::Transform findPositionOnPlan(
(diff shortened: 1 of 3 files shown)


Timple (approved):
And as a bonus, this does increase readability!

Rayman:
Google style guide (and cpplint) used to recommend output parameters to be pointer so that you can see an argument is modified:
```c++
foo(x, y);     // x and y won't be mutated
bar(x, &y);    // y may be mutated
```
But I can't find it in the style guide anymore, is this rule outdated?
https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

lewie-donckers:
> But I can't find it in the style guide anymore, is this rule outdated? 

I hope so. ;-D

Rayman (approved):
I've did some research and it seems that roslint was not enabled anymore, and that the updated guide recommends using references. This led me to make https://github.com/nobleo/path_tracking_pid/pull/78 to re-enable roslint. roslint succeeds with this PR merged into it.

PR #77, opened 2022-02-22
Enable VERBOSE_TESTS

Rayman:
If we enable `VERBOSE_TESTS`, you can see which unittests have executed. This gives confidence all the tests are running as expected.

See the logs of the action of the current PR so see how it now looks.


lewie-donckers (approved):
This gives very verbose output. But I prefer that to not seeing anything. It will also help track down problems in case any pop up.

Timple (approved):

PR #79, opened 2022-02-22
Switched distToSegmentSquared() to proper units.

lewie-donckers:
Boot Units are go! 🚀 This will be the first PR in a series. I will change a few functions and/or variables per PR. This keeps the changes small and easy to review. Gradually all conversions from (`.value()`) and to (`* units::meter`) the strong types will move to the outer edges of the code.

Changed `Controller::distToSegmentSquared()` to Boost Units.

### DIFF
--- a/include/path_tracking_pid/units.hpp
+++ b/include/path_tracking_pid/units.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <boost/units/cmath.hpp>
 #include <boost/units/systems/si.hpp>
 
 namespace path_tracking_pid::units
@@ -27,10 +28,10 @@ using acceleration_t = boost::units::quantity<boost::units::si::acceleration, do
 // auto v1 = 23.0 * square_meter;  // type of v1 is distance_squared_t
 // auto v2 = 5 * meter_per_second; // type of v1 is boost::units::quantity<boost::units::si::velocity, int>
 // ```
-using boost::units::si::meter;
-using boost::units::si::meter_per_second;
-using boost::units::si::meter_per_second_squared;
-using boost::units::si::second;
-using boost::units::si::square_meter;
+inline const auto meter = boost::units::si::meter;
+inline const auto meter_per_second = boost::units::si::meter_per_second;
+inline const auto meter_per_second_squared = boost::units::si::meter_per_second_squared;
+inline const auto second = boost::units::si::second;
+inline const auto square_meter = boost::units::si::square_meter;
 
 }  // namespace path_tracking_pid::units
--- a/include/path_tracking_pid/units.hpp
+++ b/include/path_tracking_pid/units.hpp
@@ -1,0 +1,36 @@
+#pragma once
+
+#include <boost/units/systems/si.hpp>
+
+namespace path_tracking_pid::units
+{
+
+// Distance in meter.
+using distance_t = boost::units::quantity<boost::units::si::length, double>;
+
+// Distance squared in square meter.
+using distance_squared_t = boost::units::quantity<boost::units::si::area, double>;
+
+// Duration in second.
+using duration_t = boost::units::quantity<boost::units::si::time, double>;
+
+// Velocity in meter per second.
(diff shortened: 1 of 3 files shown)


include/path_tracking_pid/units.hpp
```diff
@@ -0,0 +1,36 @@
...
+using distance_t = boost::units::quantity<boost::units::si::length, double>;
+
+// Distance squared in square meter.
+using distance_squared_t = boost::units::quantity<boost::units::si::area, double>;
+
+// Duration in second.
+using duration_t = boost::units::quantity<boost::units::si::time, double>;
+
+// Velocity in meter per second.
+using velocity_t = boost::units::quantity<boost::units::si::velocity, double>;
```
Timple, line 18: What's the difference between `boost::units::quantity<boost::units::si::velocity, double>` and `boost::units::si::velocity`?

lewie-donckers: There is no difference: double is the default value type. I just wanted to make it explicit we're using doubles.

lewie-donckers: Ah, wait. May have misunderstood the question. Boost Units makes a distinction between units of measurement (like velocity, length, etc) and quantities, which are values of a certain unit of measurement.

- `boost::units::si::velocity` is a unit. and can't be used directly for storing values or calculations.
- `boost::units::quantity<boost::units::si::velocity, double>` is a quantity. a value of type `double` with the velocity unit of measurement.

`boost::units::quantity<boost::units::si::velocity, double>` is identical to `boost::units::quantity<boost::units::si::velocity>`.

For more info, see: https://www.boost.org/doc/libs/1_78_0/doc/html/boost_units/Quick_Start.html

rokusottervanger (commented):
Can I just remark that I think we moved to the implementation of Boost Units rather swiftly. In my mind, the discussion was not just about the question "Should we go for either Boost or our own implementation?" but rather "Should we go for a strong-typed unit system, and if we do, how should we do that?". I feel like we haven't even decided on this first part of the question.

Here's a few arguments against:
 - It increases verbosity
 - It increases complexity for newbie developers

And it hasn't solved any bugs yet, if I'm not mistaken.

I agree that it is a fun (thought) exercise from a software engineering point of view, but maybe we have more pressing problems that are worth spending this time on?

lewie-donckers:
> Can I just remark that I think we moved to the implementation of Boost Units rather swiftly. In my mind, the discussion was not just about the question "Should we go for either Boost or our own implementation?" but rather "Should we go for a strong-typed unit system, and if we do, how should we do that?". I feel like we haven't even decided on this first part of the question.
> 
> Here's a few arguments against:
> 
> * It increases verbosity
> * It increases complexity for newbie developers (of which we have a few)
> 
> And it hasn't solved any bugs yet, if I'm not mistaken.
> 
> I agree that it is a fun (thought) exercise from a software engineering point of view, but maybe we have more pressing problems that are worth spending this time on?

Okay... Maybe we need a different way for me to propose changes than these PRs. I thought I had 'permission' to continue and that we were all on the same page. Clearly this was not yet the case for this change. Perhaps we should discuss certain changes in a Teams meeting? I do need regular and timely feedback, else I would be blocked.

As to the arguments against it:
- it indeed increases verbosity. imo, this is a small price to pay for making it very explicit what we're dealing with (the types and units of measurements).
- it could indeed by daunting for newbie developers. but if a project limits itself to the level of its least able developers, i predict you end up with low code quality. i suggest you set your sights a bit higher, aim for the average or preferably slightly above it. with training/guidance using `velocity_t` should not be a big leap from `double`.

Indeed, these changes have not fixed any bugs yet. But that wasn't my assignment. My assignment is to increase code quality. And with a unit of measurement library you can prevent entire classes of errors. The compiler will simply prevent you from doing so.

lewie-donckers:
Closing this PR. Boost units changes can still be seen here: https://github.com/nobleo/path_tracking_pid/compare/cleanup/boost-units-4

PR #85, opened 2022-02-23
Refactored output parameters of distToSegmentSquared().

lewie-donckers:
Refactored the output parameters of `Controller::distToSegmentSquared()` to return values.

See:
- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters
- https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f21-to-return-multiple-out-values-prefer-returning-a-struct-or-tuple

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -192,21 +192,17 @@ public:
   double getVelMaxObstacle() const;
 
 private:
-  void distToSegmentSquared(
-    const tf2::Transform & pose_p, const tf2::Transform & pose_v, const tf2::Transform & pose_w,
-    tf2::Transform & pose_projection, double & distance_to_p, double & distance_to_w) const;
-
-  // Overloaded function for callers that don't need the additional results
-  double distToSegmentSquared(
-    const tf2::Transform & pose_p, const tf2::Transform & pose_v,
-    const tf2::Transform & pose_w) const
+  // Result of distToSegmentSquared().
+  struct DistToSegmentSquaredResult
   {
-    tf2::Transform dummy_tf;
-    double dummy_double;
-    double result;
-    distToSegmentSquared(pose_p, pose_v, pose_w, dummy_tf, result, dummy_double);
-    return result;
-  }
+    tf2::Transform pose_projection;
+    double distance2_to_p = 0;  // Square distance in meter squared.
+    double distance_to_w = 0;   // Distance in meter.
+  };
+
+  DistToSegmentSquaredResult distToSegmentSquared(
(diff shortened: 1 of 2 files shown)


include/path_tracking_pid/controller.hpp
```diff
@@ -192,21 +192,17 @@ class Controller : private boost::noncopyable
...
-  void distToSegmentSquared(
-    const tf2::Transform & pose_p, const tf2::Transform & pose_v, const tf2::Transform & pose_w,
-    tf2::Transform & pose_projection, double & distance_to_p, double & distance_to_w) const;
+  // Result of distToSegmentSquared().
+  struct DistToSegmentSquaredResult
+  {
+    tf2::Transform pose_projection;
+    double distance2_to_p = 0;  // Square distance in meter squared.
+    double distance_to_w = 0;   // Distance in meter.
+  };
```
Rayman, line 201: I'm trying to figure out what `distToSegmentSquared` is doing. My guess is to calculate the closest point on a line to another point. Maybe we could document this.

But why do we need `distance2_to_p` and `distance_to_w`. If you have the closest point, you can calculate these distances pretty easily. They don't need to be return values.

lewie-donckers: > I'm trying to figure out what `distToSegmentSquared` is doing. My guess is to calculate the closest point on a line to another point. Maybe we could document this.

Sure. But I would need some help. Could you supply me with the documentation blob?

> But why do we need `distance2_to_p` and `distance_to_w`. If you have the closest point, you can calculate these distances pretty easily. They don't need to be return values.
> 
> I would choose the following signature:
> 
> ```c++
>  tf2::Transform distToSegmentSquared(const tf2::Transform & pose_p, const tf2::Transform & pose_v, const tf2::Transform & pose_w);
> ```

Excellent idea. But I propose we do that in another PR. So this PR just refactors the output parameters with as few other changes as possible.

Rayman: Maybe @cesar-lopez-mar or somebody else could correct me:
```
@brief Closest point between a line segment and a point

Calculate the closest point between the line segment bounded by PV and the point W.

p: Start of the line segment
v: End of the line segment
w: The point
```

cesar-lopez-mar: Correct. And the reason we use signal squared, is to avoid a lot of sqrt operations. We only need to find the closes position of the robot the path.

lewie-donckers: @Rayman fixed.

Rayman (approved):

PR #91, opened 2022-02-23
Refactored distToSegmentSquared().

lewie-donckers:
Refactored `Controller::distToSegmentSquared()`.

Some of the changes:
- prefer early returns over long/nested ifs: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting
- prefer const for local variables: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con1-by-default-make-objects-immutable
- removed incorrect (or outdated?) comment

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -36,14 +36,6 @@ bool have_same_sign(T val1, T val2)
   return std::signbit(val1) == std::signbit(val2);
 }
 
-// Floating-point aware version of std::clamp(). fclamp() is to std::clamp() what std::fmin() and
-// std::fmax() are to std::min() and std::max().
-template <typename T>
-T fclamp(T value, T low, T high)
-{
-  return std::fmax(low, std::fmin(value, high));
-}
-
 // Converts a pose to the corresponding transform.
 tf2::Transform to_transform(const geometry_msgs::Pose & pose)
 {
@@ -306,7 +298,7 @@ Controller::DistToSegmentSquaredResult Controller::distToSegmentSquared(
 
   DistToSegmentSquaredResult result;
 
-  const double t = fclamp(
+  const double t = std::clamp(
     ((pose_p.getOrigin().x() - pose_v.getOrigin().x()) *
        (pose_w.getOrigin().x() - pose_v.getOrigin().x()) +
      (pose_p.getOrigin().y() - pose_v.getOrigin().y()) *
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -36,6 +36,14 @@ bool have_same_sign(T val1, T val2)
   return std::signbit(val1) == std::signbit(val2);
 }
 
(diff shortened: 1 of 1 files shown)


src/controller.cpp
```diff
@@ -36,6 +36,14 @@ bool have_same_sign(T val1, T val2)
   return std::signbit(val1) == std::signbit(val2);
 }
 
+// Floating-point aware version of std::clamp(). fclamp() is to std::clamp() what std::fmin() and
+// std::fmax() are to std::min() and std::max().
+template <typename T>
+T fclamp(T value, T low, T high)
+{
+  return std::fmax(low, std::fmin(value, high));
+}
```
Rayman, line 45: Why didn't `std::clamp` not work for this case?

lewie-donckers: That depends. In the original code `fmin()` and `fmax()` were used:
```cpp
t = fmax(0.0, fmin(1.0, t));
```

These functions explicitly deal with NAN values while `min()` and `max()` do not. I assumed this was a conscious choice by the original implementer.

If so, `std::clamp()` would not suffice.

If not, then I could change to `std::clamp()`.

lewie-donckers: See https://en.cppreference.com/w/cpp/numeric/math/fmin

Rayman: Ah OK, you didn't want to change behavior of the code. But I think the original implementer didn't take paths that contain NaN values into account, so my preference is to use `std::clamp`. I'm curious what other reviewers think of this

Rayman: If paths contain NaN values, we should do an early check for that and reject the path, because I assume other paths of the code didn't take NaNs into account

rokusottervanger: > If paths contain NaN values, we should do an early check for that and reject the path, because I assume other paths of the code didn't take NaNs into account

I agree with you, Ramon, but I guess that would be for another PR, to keep this one a clean refactor. I've raised issues for a few of these behavior/functionality improvements already, but I guess there is room for more.

Timple: https://coliru.stacked-crooked.com/view?id=0b50b3c84a79727d

Seems like NAN input returns NAN output?

Rayman: Yes, but this is not the case for the `fmax` and `fmin`
> If one of the two arguments is NaN, the value of the other argument is returned
If the input is NaN, the current code returns `high` and this stays the same after the refactor

Timple: > If one of the two arguments is NaN, the value of the other argument is returned

Ah yak. I'm quite sure this should not be the case. Let me look into this briefly and we can simply use `std::clang` here instead of opening issues for it.

Timple: Verified, `std::clang` is better even.

lewie-donckers: @Timple @Rayman fixed!

Timple (approved):

PR #96, opened 2022-02-24
Extracted deltas_of_plan().

lewie-donckers:
Extracted pure function `deltas_of_plan()` from `Controller::setPlan()` and added some unit tests.


Timple:
As mentioned in #99 I think this pr should be combined with that one.

Timple (approved):

Timple:
Approved this one, since it's used in two places.

PR #99, opened 2022-02-24
Extracted inverse_turning_radiuses().

lewie-donckers:
Extracted pure function `inverse_turning_radiuses()` from `Controller::setPlan()` and added some unittests.

### DIFF
--- a/src/calculations.cpp
+++ b/src/calculations.cpp
@@ -40,4 +40,25 @@ std::vector<double> distances_to_goal(const std::vector<tf2::Transform> & deltas
   return result;
 }
 
+std::vector<double> inverse_turning_radiuses(const std::vector<tf2::Transform> & deltas)
+{
+  auto result = std::vector<double>{};
+
+  result.reserve(deltas.size() + 1);
+  std::transform(
+    deltas.cbegin(), deltas.cend(), std::back_inserter(result), [](const tf2::Transform & d) {
+      const auto & origin = d.getOrigin();
+      const auto dpX = origin.x();
+      const auto dpY = origin.y();
+      const auto dpXY2 = std::pow(dpX, 2) + std::pow(dpY, 2);
+      if (dpXY2 < FLT_EPSILON) {
+        return std::numeric_limits<double>::infinity();
+      }
+      return (2 * dpY) / dpXY2;
+    });
+  result.push_back(0.0);
+
+  return result;
+}
+
 }  // namespace path_tracking_pid
--- a/src/calculations.cpp
+++ b/src/calculations.cpp
@@ -1,6 +1,7 @@
 #include "calculations.hpp"
 
 #include <algorithm>
+#include <limits>
 #include <numeric>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


src/calculations.cpp
```diff
@@ -40,4 +40,25 @@ std::vector<double> distances_to_goal(const std::vector<tf2::Transform> & deltas
   return result;
 }
 
+std::vector<double> inverse_turning_radiuses(const std::vector<tf2::Transform> & deltas)
+{
+  auto result = std::vector<double>{};
+
+  result.reserve(deltas.size() + 1);
```
Timple, line 47: Hah, didn't know this one yet.

src/calculations.cpp
```diff
@@ -40,4 +40,25 @@ std::vector<double> distances_to_goal(const std::vector<tf2::Transform> & deltas
...
+  std::transform(
+    deltas.cbegin(), deltas.cend(), std::back_inserter(result), [](const tf2::Transform & d) {
```
Timple, line 49: What's the benefit of this construction versus:

```c++
for (const auto delta : deltas) {
  radius = magic;
  result.push_back(radius)
}
return result;

```

test/unittests/test_calculations.cpp
```diff
@@ -84,4 +85,51 @@ TEST(PathTrackingPidCalculations, DistancesToGoal_IgnoreZ)
...
+TEST(PathTrackingPidCalculations, InverseTurningRadiuses_Empty)
+{
+  const auto deltas = std::vector<tf2::Transform>{};
+  const auto ref = std::vector<double>{0.0};
+  const auto result = inverse_turning_radiuses(deltas);
+
+  EXPECT_EQ(ref, result);
+}
+
+TEST(PathTrackingPidCalculations, InverseTurningRadiuses_SingleTooSmall)
```
Timple, line 97: Thanks for adding tests!

src/controller.cpp
```diff
@@ -228,23 +228,7 @@ void Controller::setPlan(
   const auto deltas = deltas_of_plan(global_plan_tf_);
```
Timple, line 228: I think it makes more sense to combine these two functions.

` turning_radius_inv_vector_ = inverse_turning_radiuses(global_plan_tf_)`;

This reads better. That you need delta's is an implementation detail not relevent for this part of the code.

lewie-donckers: I could do that but we also need the deltas for the distances to the goal (` distance_to_goal_vector_ = distances_to_goal(deltas);`). See PR #98 . If i merge the deltas into those functions, we'd have to do that work twice.

Timple: Well, you can call that function as first step in this function right?

No duplicate code in that case.

lewie-donckers: No duplicate code. But now `deltas_of_plan()` is called once in `Controller::setPlan()`:

```cpp
  const auto deltas = deltas_of_plan(global_plan_tf_);

  distance_to_goal_vector_ = distances_to_goal(deltas);
  turning_radius_inv_vector_ = inverse_turning_radiuses(deltas);
```

If we inline `deltas_of_plan()` in `distances_to_goal()` and `inverse_turning_radiuses()`, that work would get done twice. I'm not saying we shouldn't, it's just something to be aware of.

Timple: Ah right, let's keep it like this!

Sorry, I was only looking at the delta :)

Timple (approved):

Rayman (approved):

PR #101, opened 2022-02-24
Refactored setPlan().

### LINKED ISSUE #97: Refactor geometry_msgs::* to tf2:: datatypes where possible in controller.hpp
- [ ] convert member variable `geometry_msgs::Transform` tf_base_to_steered_wheel_
- [ ] convert all `geometry_msgs::Transform` that are used as arguments
  - [ ]  maybe start with `setPlan`, move the conversion code to the ROS wrapper
  - [ ] `global_plan_` doesn't need to be a member

lewie-donckers:
Next PR to fix #97 : refactored `Controller::setPlan()` to use tf2 types.

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -37,57 +37,43 @@ bool have_same_sign(T val1, T val2)
   return std::signbit(val1) == std::signbit(val2);
 }
 
-// Converts a pose to the corresponding transform.
-tf2::Transform to_transform(const geometry_msgs::Pose & pose)
-{
-  tf2::Transform result;
-  tf2::fromMsg(pose, result);
-  return result;
-}
-
 // Returns the square distance between two points
 double distSquared(const tf2::Transform & a, const tf2::Transform & b)
 {
   return a.getOrigin().distance2(b.getOrigin());
 }
 
-// Return the square distance between two points.
-double distSquared(const geometry_msgs::Pose & a, const geometry_msgs::Pose & b)
-{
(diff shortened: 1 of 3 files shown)


src/controller.cpp
```diff
@@ -172,14 +158,10 @@ TricycleSteeringCmdVel Controller::computeTricycleModelInverseKinematics(
 }
 
 void Controller::setPlan(
-  const geometry_msgs::Transform & current_tf, const geometry_msgs::Twist & odom_twist,
-  const std::vector<geometry_msgs::PoseStamped> & global_plan)
+  const tf2::Transform & current_tf, const geometry_msgs::Twist & odom_twist,
+  const std::vector<tf2::Transform> & global_plan)
 {
-  ROS_DEBUG("TrackingPidLocalPlanner::setPlan(%d)", (int)global_plan.size());
-  ROS_DEBUG("Plan is defined in frame '%s'", global_plan.at(0).header.frame_id.c_str());
```
Rayman, line 179: Where did this log message go?

lewie-donckers: I removed it. Since it was only debug logging, it didn't seem that important. The information to log this is no longer available (here).

If you want I can move the logging to `TrackingPidLocalPlanner`.

Rayman: Ah yes because Transforms don't contain have a header

Rayman (approved):

Timple (approved):

PR #105, opened 2022-02-24
Removed global_plan_ member.

### LINKED ISSUE #97: Refactor geometry_msgs::* to tf2:: datatypes where possible in controller.hpp

lewie-donckers:
Last PR to fix #97 - removed `TrackingPidLocalPlanner::global_plan_` member variable.

Please check this one carefully. @Rayman _thought_ not republishing the plan in `TrackingPidLocalPlanner::reconfigure_pid()` would be ok, but wasn't entirely sure.

### DIFF
--- a/include/path_tracking_pid/path_tracking_pid_local_planner.hpp
+++ b/include/path_tracking_pid/path_tracking_pid_local_planner.hpp
@@ -123,7 +123,6 @@ private:
   ros::Duration prev_dt_;
   path_tracking_pid::Controller pid_controller_;
 
-  std::vector<geometry_msgs::PoseStamped> global_plan_;
   nav_msgs::Path received_path_;
 
   // Obstacle collision detection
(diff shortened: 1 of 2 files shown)


Rayman (approved):

Timple:
I'm okay without the republish. As we shouldn't modify the global plan.

But I don't like the `global_plan_copy` name. Because it's only a copy for a very brief moment. Because then it's modified and not a copy anymore.

How about `global_plan_map_frame`? Because that's what we're achieving.

lewie-donckers:
> global_plan_map_frame

@Timple fixed!

Timple (approved):

rokusottervanger (approved):

Issue #82, opened 2022-02-22
Simplify the math being done to calculate direction to goal


rokusottervanger:
A simple dot product would give the desired outcome: the length of the projection of one vector onto the other, with the sign of the cosine of the angle between the two vectors (which is what we're looking for).

https://github.com/nobleo/path_tracking_pid/blob/61c6cc828bd98dddd8cc14cbabdd62353a89ad3c/src/controller.cpp#L581-L593

PR #108, opened 2022-02-28
Simplified the calculation of goal direction.

lewie-donckers:
Simplified the calculation of the direction of the goal in `Controller::update()`. Fix for #82 .

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -523,7 +523,7 @@ Controller::UpdateResult Controller::update(
   ROS_DEBUG("distance_to_goal: %f", distance_to_goal_);
 
   const auto in_direction_of_goal =
-    !std::signbit(current_tf.getOrigin().dot(current_goal_.getOrigin()));
+    is_in_direction_of_target(current_tf, current_goal_.getOrigin(), target_x_vel);
 
   // If we are as close to our goal or closer then we need to reach end velocity, enable end_phase.
   // However, if robot is not facing to the same direction as the local velocity target vector, don't enable end_phase.
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -30,13 +30,6 @@ constexpr double ang_lower_limit = -100.0;
 // Anti-windup term. Limits the absolute value of the integral term.
 constexpr double windup_limit = 1000.0;
 
-// Indicates if both values have the same sign.
-template <typename T>
-bool have_same_sign(T val1, T val2)
-{
-  return std::signbit(val1) == std::signbit(val2);
-}
-
 // Returns the square distance between two points
 double distSquared(const tf2::Transform & a, const tf2::Transform & b)
 {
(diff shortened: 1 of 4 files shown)


Timple (approved):

Rayman (approved):

lewie-donckers:
@Rayman @rokusottervanger @Timple 
I need some help. I noticed that I introduced a change in behavior which can be seen when running the visualization of the tests. I tracked it down to this commit.

In certain tests, you can see that the robot takes longer to stop and turns a bit more at the end of the test.

Example before change:
image

Example after change:
image

Any idea what's going on and how to fix it?

rokusottervanger:
I'm not sure. But could you maybe refactor this first, so that this calculation is isolated as a separate function. Then a unit test can show if/how the behavior changes.

src/controller.cpp
```diff
@@ -529,14 +522,13 @@ Controller::UpdateResult Controller::update(
...
-  tf2::Transform base_to_goal = current_tf.inverseTimes(current_goal_);
-  const double angle_to_goal = atan2(base_to_goal.getOrigin().x(), -base_to_goal.getOrigin().y());
+  const auto in_direction_of_goal =
+    !std::signbit(current_tf.getOrigin().dot(current_goal_.getOrigin()));
 
   // If we are as close to our goal or closer then we need to reach end velocity, enable end_phase.
   // However, if robot is not facing to the same direction as the local velocity target vector, don't enable end_phase.
   // This is to avoid skipping paths that start with opposite velocity.
-  if ((distance_to_goal_ <= fabs(d_end_phase)) && have_same_sign(target_x_vel, angle_to_goal)) {
+  if ((distance_to_goal_ <= fabs(d_end_phase)) && in_direction_of_goal) {
```
rokusottervanger, line 531: Where's the comparison with the target_x_vel gone?

src/controller.cpp
```diff
@@ -529,14 +522,13 @@ Controller::UpdateResult Controller::update(
   ROS_DEBUG("d_end_phase: %f", d_end_phase);
   ROS_DEBUG("distance_to_goal: %f", distance_to_goal_);
 
-  // Get 'angle' towards current_goal
-  tf2::Transform base_to_goal = current_tf.inverseTimes(current_goal_);
-  const double angle_to_goal = atan2(base_to_goal.getOrigin().x(), -base_to_goal.getOrigin().y());
+  const auto in_direction_of_goal =
+    !std::signbit(current_tf.getOrigin().dot(current_goal_.getOrigin()));
```
rokusottervanger, line 527: I like the seeming simplicity of this, but `current_tf` and `current_goal_` are both defined in the `map` frame, so you're now projecting the vector from `map` to `current_tf` onto the vector from `map` to `current_goal_`.  This doesn't do the same as in the original calculation.

rokusottervanger (changes requested):

src/controller.cpp
```diff
@@ -529,14 +522,13 @@ Controller::UpdateResult Controller::update(
...
```
rokusottervanger, line 526: I was thinking something along the lines of this:
```suggestion
const auto base_to_next_path_pose = current_goal_.getOrigin() - current_tf.getOrigin();
  const auto in_direction_of_goal =
    !std::signbit(current_goal_.getBasis().tdotx(base_to_next_path_pose)*target_x_vel);
```
But I'm not sure if it's entirely correct. I now project onto the current_goal_ x vector, but one could also pick the current_tf x vector. Not sure which is correct, and how to test this in an integration test.

Also, I was inspired by your `!std::signbit`, and added the multiplication with the current velocity setpoint. That doesn't make it any more transparent...

lewie-donckers: I'm not sure if it's correct or not but it did give different results than the original implementation (for certain inputs). Based on your remarks, I changed it to the following: 

```cpp
  const auto base_to_next_path_pose = current_goal_.getOrigin() - current_tf.getOrigin();
  const auto in_direction_of_goal =
    !std::signbit(current_tf.getBasis().tdotx(base_to_next_path_pose) * target_x_vel);
```

This does give the same results as the original implementation for all inputs that occur in the automated tests.

I plan to do the following:
- move the new implementation to a named function in `calculations.h/cpp`
- add some unit tests

I could use some help with the documentation and coming up with interesting inputs for the unit tests

lewie-donckers:
@Timple @rokusottervanger @Rayman 

I have reworked this PR. I fixed the new implementation with the help of @rokusottervanger . The calculation is now moved to a separate function and I've added some unit tests.

I tested this by running a build with both the old and new calculation in tandem. It logged if they were not the same. None of the automated tests showed any differences.

Timple (approved):

rokusottervanger (approved):

Issue #90, opened 2022-02-23
Controller inherits from `nav_core::BaseLocalPlanner`, but does not support it


rokusottervanger:
The controller as a child of `nav_core::BaseLocalPlanner` has not been tested for a long time. It has been decided to drop this support (see #89 ), so we should remove the inheritance here:
https://github.com/nobleo/path_tracking_pid/blob/61c6cc828bd98dddd8cc14cbabdd62353a89ad3c/src/path_tracking_pid_local_planner.cpp#L22-L25

PR #110, opened 2022-02-28
Disable nav_core::BaseLocalPlanner support.

lewie-donckers:
Fix for #90 : disabled the `nav_core::BaseLocalPlanner` support.


Timple (changes requested):
We should not expose this as a plugin anymore in that case: https://github.com/nobleo/path_tracking_pid/blob/main/path_tracking_pid_plugin.xml

Timple (approved):

PR #111, opened 2022-02-28
Controller no longer changes plan.

### LINKED ISSUE #86: Global plan is changed to the controller's liking
IMO, if the controller is incapable of handling the global plan, it should reject the global plan, not alter it to its liking. This `filter_plan` call does exactly that. I think we should just check it, and return `REJECTED` if it does not satisfy the controller's constraints to the global plan.

https://github.com/nobleo/path_tracking_pid/blob/61c6cc828bd98dddd8cc14cbabdd62353a89ad3c/src/controller.cpp#L183-L188

lewie-donckers:
`Controller` no longer changes the plan. It now checks it and rejects it if it doesn't like it. This error is propagated back to ROS. Fixes #86 .

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -43,33 +43,27 @@ bool is_pose_angle_obtuse(
   return distSquared(prev, next) > (distSquared(prev, cur) + distSquared(cur, next));
 }
 
-// Filters the given plan. The first and last poses are always accepted (if they exist).
-// Intermediate poses are only accepted if the angle (with respect to the previous and next poses)
-// is obtuse.
-std::vector<tf2::Transform> filter_plan(const std::vector<tf2::Transform> & plan)
+/**
+ * Checks the given plan. The first and last poses are always accepted (if they exist). Intermediate
+ * poses are only accepted if the angle (with respect to the previous and next poses) is obtuse.
+ * 
+ * @param[in] plan Plan to check.
+ * @return True if all poses in the plan are accepted. False otherwise.
+ */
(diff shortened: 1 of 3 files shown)


Rayman:
Nice work!

Since this changes the behavior of the controller, should we do an integration test on harvey?

src/controller.cpp
```diff
@@ -150,19 +144,19 @@ TricycleSteeringCmdVel Controller::computeTricycleModelInverseKinematics(
...
 {
   ROS_DEBUG("TrackingPidLocalPlanner::setPlan(%zu)", global_plan.size());
 
-  global_plan_tf_ = filter_plan(global_plan);
-  if (global_plan_tf_.size() != global_plan.size()) {
-    ROS_WARN(
-      "Not all poses of path are used since not all poses were in the expected direction of the "
-      "path!");
+  if (!check_plan(global_plan)) {
+    ROS_WARN("Rejected plan because not all poses were in the expected direction of the path!");
```
Timple, line 154: This can be a `ROS_ERROR`. As we don't continue

Rayman:
I've search for that specific error message in the bagfiles that I have on my laptop. I couldn't find it so harvey should not be affected by this change
```
time grepros 'key=Not all poses of path are used' --topic /rosout *

real	6m40.212s
user	6m8.666s
sys	0m10.115s
```
I'm now running it on all bagfiles of February, just to be sure

Timple (approved):
Thank you for the extra warning. We concluded that this should not occur in production.

Rayman (approved):

PR #115, opened 2022-03-01
Refactored output parameter.

lewie-donckers:
Refactored `TrackingPidLocalPlanner::computeVelocityCommands()` to use a return value instead of an output parameter.

The function never read from the output parameter (it was pure output and not in-/output), and the results were one of the following:
- false and the output parameter was not set
- true and the output parameter was set

This maps perfectly to `std::optional`. Either the function is successful and it returns the velocity command or it is not successful and it does not return a velocity command.

### DIFF
--- a/include/path_tracking_pid/path_tracking_pid_local_planner.hpp
+++ b/include/path_tracking_pid/path_tracking_pid_local_planner.hpp
@@ -90,10 +90,9 @@ public:
 private:
   /**
    * @brief Calculates the velocity command based on the current robot pose given by pose.
-   * @param cmd_vel Output the velocity command
-   * @return true if succeded.
+   * @return Velocity command on success, empty optional otherwise.
    */
-  bool computeVelocityCommands(geometry_msgs::Twist & cmd_vel);
+  std::optional<geometry_msgs::Twist> computeVelocityCommands();
 
   /**
    * @brief Returns true, if the goal is reached.
--- a/src/path_tracking_pid_local_planner.cpp
+++ b/src/path_tracking_pid_local_planner.cpp
@@ -209,7 +209,7 @@ bool TrackingPidLocalPlanner::setPlan(const std::vector<geometry_msgs::PoseStamp
   return true;
 }
 
-bool TrackingPidLocalPlanner::computeVelocityCommands(geometry_msgs::Twist & cmd_vel)
+std::optional<geometry_msgs::Twist> TrackingPidLocalPlanner::computeVelocityCommands()
 {
   ros::Time now = ros::Time::now();
   if (prev_time_.isZero()) {
@@ -219,14 +219,14 @@ bool TrackingPidLocalPlanner::computeVelocityCommands(geometry_msgs::Twist & cmd
   if (dt.isZero()) {
     ROS_ERROR_THROTTLE(
       5, "dt=0 detected, skipping loop(s). Possible overloaded cpu or simulating too fast");
-    cmd_vel = geometry_msgs::Twist();
+    auto cmd_vel = geometry_msgs::Twist();
     cmd_vel.linear.x = pid_controller_.getControllerState().current_x_vel;
     cmd_vel.angular.z = pid_controller_.getControllerState().current_yaw_vel;
-    return true;  // False is no use: https://github.com/magazino/move_base_flex/issues/195
+    return cmd_vel;
   }
   if (dt < ros::Duration(0) || dt > ros::Duration(DT_MAX)) {
     ROS_ERROR("Invalid time increment: %f. Aborting", dt.toSec());
-    return false;
+    return std::nullopt;
   }
   try {
     ROS_DEBUG("map_frame: %s, base_link_frame: %s", map_frame_.c_str(), base_link_frame_.c_str());
@@ -234,7 +234,7 @@ bool TrackingPidLocalPlanner::computeVelocityCommands(geometry_msgs::Twist & cmd
   } catch (const tf2::TransformException & ex) {
     ROS_ERROR("Received an exception trying to transform: %s", ex.what());
     active_goal_ = false;
-    return false;
+    return std::nullopt;
   }
 
   // Handle obstacles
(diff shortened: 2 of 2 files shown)


src/path_tracking_pid_local_planner.cpp
```diff
@@ -215,22 +215,22 @@ bool TrackingPidLocalPlanner::computeVelocityCommands(geometry_msgs::Twist & cmd
...
+    return std::nullopt;
   }
   try {
     ROS_DEBUG("map_frame: %s, base_link_frame: %s", map_frame_.c_str(), base_link_frame_.c_str());
     tfCurPoseStamped_ = tf_->lookupTransform(map_frame_, base_link_frame_, ros::Time(0));
   } catch (const tf2::TransformException & ex) {
     ROS_ERROR("Received an exception trying to transform: %s", ex.what());
     active_goal_ = false;
-    return false;
+    return std::nullopt;
```
Rayman, line 233: The MBF interface that we implement supports returning specific error codes for several of these failure modes. For example, and `TransformException` could return `uint8 TF_ERROR = 111` to indicate a transformation went wrong. This way the caller can figure out what went wrong.

I think `std::optional` doesn't really map well onto that. The command velocity is not optional. It is required, but it can go wrong due to various errors. I think raising exceptions for these errors would be better than the current implementation.

lewie-donckers: Please be aware that I did not change the behavior of the code, I just refactored an output parameter. Based on the behavior of the previous implementation, `std::optional` is a perfect fit. The velocity command _is_ optional, because the operation can fail.

However, if you want to improve the code further and have this helper function also return the reason for the failure (using the error code you mentioned), I suggest we switch to a type from the Boost Outcome library, like `boost::checked<T, E>` which closely resembles the proposed `std::expected`. It is a type which represents a value of type T on success or an error value of type E on failure. But I think that should be done in a separate PR.

lewie-donckers: Just to sure we're talking about the same thing here: the overload of `TrackingPidLocalPlanner::computeVelocityCommands()` I refactored in this PR, is no longer an interface function. We stopped inheriting from `nav_core::BaseLocalPlanner` in another PR. So now this is just a private helper function.

src/path_tracking_pid_local_planner.cpp
```diff
@@ -215,22 +215,22 @@ bool TrackingPidLocalPlanner::computeVelocityCommands(geometry_msgs::Twist & cmd
   if (dt.isZero()) {
     ROS_ERROR_THROTTLE(
       5, "dt=0 detected, skipping loop(s). Possible overloaded cpu or simulating too fast");
-    cmd_vel = geometry_msgs::Twist();
+    auto cmd_vel = geometry_msgs::Twist();
     cmd_vel.linear.x = pid_controller_.getControllerState().current_x_vel;
     cmd_vel.angular.z = pid_controller_.getControllerState().current_yaw_vel;
-    return true;  // False is no use: https://github.com/magazino/move_base_flex/issues/195
+    return cmd_vel;
```
Rayman, line 221: Can you re-add this comment, it provides some useful information.
```
// At the first call of computeVelocityCommands we can't calculate a cmd_vel. We could elaborate on this:
// We can't return false because of https://github.com/magazino/move_base_flex/issues/195 so the current velocity is send instead.
```

src/path_tracking_pid_local_planner.cpp
Rayman: I'm not a fan of this approach, but I won't block this PR for that. I think we should definitely focus on the error handling part more in the future. `std::expected` or `boost::checked` is a good tip

Rayman (approved):
I'm not a fan of this approach, but I won't block this PR for that. I think we should definitely focus on the error handling part more in the future. std::expected or boost::checked is a good tip

include/path_tracking_pid/path_tracking_pid_local_planner.hpp
```diff
@@ -90,10 +90,9 @@ class TrackingPidLocalPlanner : public mbf_costmap_core::CostmapController,
 private:
   /**
    * @brief Calculates the velocity command based on the current robot pose given by pose.
-   * @param cmd_vel Output the velocity command
-   * @return true if succeded.
+   * @return Velocity command on success, empty optional otherwise.
    */
-  bool computeVelocityCommands(geometry_msgs::Twist & cmd_vel);
```
Timple, line 96: This whole function only existed for the sake of the `nav_core` implementation.

Now that we don't have that legacy anymore. I think there are much better improvements to be made.

No we're calling `computeVelocityCommands` from the `computeVelocityCommands` function. It doesn't add any semantics. The 'main' `computeVelocityCommands` could be broken into functions with better descriptions.

lewie-donckers: I agree there are more and better improvements to be made. Nonetheless, this is also an improvement.  It may be small, but it's an improvement - at least from my perspective.

The investment has already been made and this won't block any other improvements from being made.

lewie-donckers:
I'm a bit puzzled. I created what I think was a very innocuous PR which is in-line with a best practice which is widely supported within the C++ community.

Yet, for such a small change, I get quite a bit of pushback from both @Rayman and @Timple . So I was hoping you two would explain to me what you don't like about this. So I can perhaps address your concerns or refrain from these kind of changes in the future.

Is it the principle of no output parameters itself, the usage of `std::optional`, something else?

Timple:
PR comments often sound a bit harsh as good stuff is rarely complemented :slightly_smiling_face: . I actually liked to learn about the `std::optional`.

My comment was that you are refactoring a function which doesn't describe it's actual use. But these kind of comments can simply be split-off to issues and the PR can continue: https://github.com/nobleo/path_tracking_pid/issues/120

Timple (approved):

Rayman:
> So I was hoping you two would explain to me what you don't like about this.

I think we were hoping and suggesting some bigger changes to the error handling that are needed, while you made a small improvement to existing code. I sometimes use a review to comment on code that surrounds the block of code because that's the first time I'm really looking at it. I shouldn't do that and focus more on the current review.

PR #118, opened 2022-03-02
Refactor collision visualization functions

Rayman:
I did an alternative refactor to #107.

- I've split the main function up in several small functions
- I've moved `ros::Time::now()` calls away from the visualization class
- I've used `tf2` datatypes if possible
- I've published everything on a single visualization topic

### DIFF
--- a/src/path_tracking_pid_local_planner.cpp
+++ b/src/path_tracking_pid_local_planner.cpp
@@ -316,8 +316,8 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
   auto projected_controller_state = pid_controller_.getControllerState();
 
   // Step until lookahead is reached, for every step project the pose back to the path
-  std::vector<geometry_msgs::Point> step_points;
-  std::vector<geometry_msgs::Point> poses_on_path_points;
+  std::vector<tf2::Vector3> step_points;
+  std::vector<tf2::Vector3> poses_on_path_points;
   std::vector<tf2::Transform> projected_steps_tf;
   auto projected_step_tf = tf2_convert<tf2::Transform>(tfCurPoseStamped_.transform);
   projected_steps_tf.push_back(projected_step_tf);  // Evaluate collision at base_link
@@ -331,15 +331,11 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
     projected_steps_tf.push_back(projected_step_tf);
 
     // Fill markers:
-    geometry_msgs::Point mkStep;
-    tf2::toMsg(next_straight_step_tf.getOrigin(), mkStep);
-    step_points.push_back(mkStep);
-    geometry_msgs::Point mkPointOnPath;
-    tf2::toMsg(projected_step_tf.getOrigin(), mkPointOnPath);
-    poses_on_path_points.push_back(mkPointOnPath);
+    step_points.push_back(next_straight_step_tf.getOrigin());
+    poses_on_path_points.push_back(projected_step_tf.getOrigin());
   }
 
-  std::vector<geometry_msgs::Point> collision_footprint_points;
+  std::vector<tf2::Vector3> collision_footprint_points;
   polygon_t previous_footprint_xy;
   polygon_t collision_polygon;
   for (const auto & projection_tf : projected_steps_tf) {
@@ -366,8 +362,8 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
     // Add footprint to marker
     geometry_msgs::Point previous_point = footprint.back();
     for (const auto & point : footprint) {
-      collision_footprint_points.push_back(previous_point);
-      collision_footprint_points.push_back(point);
+      collision_footprint_points.push_back(tf2_convert<tf2::Vector3>(previous_point));
+      collision_footprint_points.push_back(tf2_convert<tf2::Vector3>(point));
       previous_point = point;
     }
   }
@@ -392,7 +388,7 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
   costmap2d->convexFillCells(collision_polygon_hull_map, cells_in_polygon);
 
   // Get the max cost inside the concave polygon
-  geometry_msgs::Point collision_point;
+  tf2::Vector3 collision_point;
   uint8_t max_cost = 0.0;
   for (const auto & cell_in_polygon : cells_in_polygon) {
     // Cost checker is cheaper than polygon checker, so lets do that first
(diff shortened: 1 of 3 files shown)


lewie-donckers:
> I did an alternative refactor to #107.
> 
> * I've split the main function up in several small functions
> * I've moved `ros::Time::now()` calls away from the visualization class
> * I've used `tf2` datatypes if possible
> * I've published everything on a single visualization topic

I like all of those changes over my refactoring. I didn't know we could publish them on the same publisher or publish them one by one. But this is more in-line with the existing implementation. I focused on non-functional changes only.

src/visualization.cpp
```diff
@@ -50,101 +49,130 @@ void Visualization::publishPlanPoint(const std_msgs::Header & header, const tf2:
...
-
-  visualization_msgs::MarkerArray mkCollision;
-  mkCollisionIndicator.scale.z = cost / 255.0;
-  mkCollisionIndicator.color.a = cost / 255.0;
-  mkCollisionIndicator.pose.position = point;
-  mkCollisionIndicator.pose.position.z = mkCollisionIndicator.scale.z * 0.5;
-  if (mkCollisionIndicator.scale.z > std::numeric_limits<float>::epsilon()) {
-    mkCollisionIndicator.action = visualization_msgs::Marker::ADD;
+  visualization_msgs::Marker marker;
+  marker.header = header;
```
lewie-donckers, line 56: I do think  you missed out on an opportunity to reduce code duplication and magic values.

Personally, I really prefer the helper functions from my refactoring to create scales/colors/markers. And to create constants for colors/scales to reduce magic values.

src/visualization.cpp
```diff
@@ -50,101 +49,130 @@ void Visualization::publishPlanPoint(const std_msgs::Header & header, const tf2:
...
+  marker.ns = "Collision object";
+  marker.pose.orientation.w = 1.0;
+  marker.id = __COUNTER__;
```
lewie-donckers, line 59: As far as I can tell, the `__COUNTER__` trick is not required. The combination of `ns` and `id` is supposed to be unique. We already have unique `ns` values.

src/visualization.cpp
```diff
@@ -50,101 +49,130 @@ void Visualization::publishPlanPoint(const std_msgs::Header & header, const tf2:
...
+  marker.type = visualization_msgs::Marker::POINTS;
+  marker.scale.x = 0.5;
+  marker.scale.y = 0.5;
+  marker.color.r = 1.0;
+  marker.color.g = 0.5;
+  marker.color.a = 1.0;
+
+  std::transform(
+    steps.begin(), steps.end(), std::back_inserter(marker.points), [](const auto & step) {
```
lewie-donckers, line 95: You use the same lambda multiple times, consider creating a simple function (in an anonymous namespace at the top of the source file) and pass that to `std::transform()`.

lewie-donckers: And/or create a function for the entire container conversion:

```cpp
auto convert_points(const std::vector<tf2::Vector3>& points)
{
  ...
  std::transform(...);
  ...
}
```

Rayman: Fixed, I've create a utility function for this conversion

src/visualization.cpp
Rayman: I've thought about this but I could not find a lot of duplicate code. Each markers color and scale is tuned by hand for visibility. Right now all values are local to the function so it's very readable.

src/path_tracking_pid_local_planner.cpp
```diff
@@ -413,14 +409,20 @@ uint8_t TrackingPidLocalPlanner::projectedCollisionCost()
   }
 
   // Fiddle the polygon into a marker message
-  std::vector<geometry_msgs::Point> collision_hull_points;
+  std::vector<tf2::Vector3> collision_hull_points;
   for (const geometry_msgs::Point point : collision_polygon) {
-    collision_hull_points.push_back(point);
+    tf2::Vector3 point_tf2;
+    tf2::fromMsg(point, point_tf2);
+    collision_hull_points.push_back(point_tf2);
```
lewie-donckers, line 416: Consider using `tf2_convert()`.

src/visualization.cpp
lewie-donckers: > I've thought about this but I could not find a lot of duplicate code. 

The following construct is repeated quite often:
```cpp
  visualization_msgs::Marker marker;
  marker.header = header;
  marker.ns = "extrapolated poses";
  marker.action = visualization_msgs::Marker::ADD;
  marker.pose.orientation.w = 1.0;
  marker.type = visualization_msgs::Marker::POINTS;
  marker.scale.x = 0.5;
  marker.scale.y = 0.5;
  marker.color.r = 1.0;
  marker.color.g = 0.5;
  marker.color.a = 1.0;
  marker.points = to_msg(steps);
  marker_pub_.publish(marker);
```

In my opinion, having a function return a marker to which you only pass the stuff that changes is better. Some advantages:
- the reader no longer has to figure out if the blocks are actually the same or not. its obvious
- if we decide we need to set/remove/change a data member we change the function and its impossible to forget a block
- its more obvious in which the blocks differ because we only have to check the arguments

> Each markers color and scale is tuned by hand for visibility. 

That's fine. It's not the actual values I have a problem with.

> Right now all values are local to the function so it's very readable.

I disagree. I find it hard to read. Because I see a lot of code which is very similar but not exactly. I suggest that you at least get rid of the magic values for colors and preferably also for scales.

If you have this
```cpp
  marker.color.g = 0.5;
  marker.color.a = 1.0;
```
The reader could wonder if green really was intended or that perhaps the r&b parts were forgotten. If you write the following (or something similar), it's immediately clear. Same is true for scales, IMHO.
```cpp
static const auto dark_green = ...{0,  0.5, 0, 1};
...
 marker.color = dark_green;
```

Rayman: I did the color name refactor as you asked. That helps a bit for readability.

For the rest of the marker, the blocks do look similar but they are different. That's why all those lines are there (doesn't violate DRY). I think this is the cleanest code we could get here. I think this
```c++
visualization_msgs::Marker marker;
marker.header = header;
marker.ns = "Collision object";
marker.type = visualization_msgs::Marker::CYLINDER;
marker.scale.x = 0.5;
marker.scale.y = 0.5;
marker.scale.z = cost / 255.0;
marker.color = red;
marker.color.a = cost / 255.0;
```
is easier to read then
```c++
auto marker_indicator = create_marker(
  header, "Collision object", visualization_msgs::Marker::CYLINDER,
  create_scale(0.5, 0.5, cost / 255.0),
  create_color(1.0, 0.0, 0.0, static_cast<float>(cost) / 255.0F), {});
```
because you don't immediately see what each argument does. Maybe this is personal preference though. With python you have keyword arguments which makes the caller side more clean, but we don't have that in C++ unfortunately. Is this good enough?

src/path_tracking_pid_local_planner.cpp
Rayman: Well, `tf2_convert` calls `tf2::fromMsg` which does the conversion the other way around. Do we want to create two conversion functions for this? `from_msg` and `to_msg`?

lewie-donckers: Oops. I do think that would be better but that doesn't need to be part of this PR.

lewie-donckers (approved):

PR #119, opened 2022-03-02
Fix lowpass & refactor filter structure

Rayman:
- Fixed second order lowpass
- Made the crossover frequency and damping tuneble
- Use the filtered signal for PID calculations
- Add a derivative filter

How the lowpass filter was derived: https://github.com/nobleo/path_tracking_pid/blob/fix/lowpass/doc/second_order_lowpass_tustin.ipynb

PidDebug output with lowpass crossover frequency at 20, 4 and 1 Hz.
lowpass_20hz
lowpass_4hz
lowpass_1hz

### DIFF
--- a/test/test_path_tracking_pid.py
+++ b/test/test_path_tracking_pid.py
@@ -104,7 +104,7 @@ class TestPathTrackingPID(unittest.TestCase):
             rospy.sleep(10.0)
             self.assertTrue(checker.slowed_down)
 
-        finished_in_time = client.wait_for_result(timeout=rospy.Duration(60))
+        finished_in_time = client.wait_for_result(timeout=rospy.Duration(120))
         self.assertTrue(finished_in_time, msg="Action call didn't return in time")
         self.assertEqual(client.get_state(), outcome_exp, msg="Wrong action outcome")
 


second_order_lowpass_tustin.ipynb
```diff
@@ -0,0 +1,275 @@
+{
```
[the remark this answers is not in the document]
Rayman: Where do you want it?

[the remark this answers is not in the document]
Rayman: I've added it because it explains how the filter was derived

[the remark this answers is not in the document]
Timple: In the `doc/lowpass` folder or something?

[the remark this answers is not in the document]
Rayman: I've moved the file to doc

cfg/Pid.cfg
```diff
@@ -33,6 +34,9 @@ gen.add("Kp_ang", double_t, 0, "Kp Angular", 1, 0, 10)
 gen.add("Ki_ang", double_t, 0, "Ki Angular", 0, 0, 2)
 gen.add("Kd_ang", double_t, 0, "Kd Angular", 0.3, 0, 10)
 
+gen.add("lowpass_cutoff", double_t, 0, "Lowpass cutoff (Hz)", 20, 0, 1000)
+gen.add("lowpass_damping", double_t, 0, "Lowpass damping", sqrt(2), 0, 10)
```
Rayman, line 38: What default to we want for `lowpass_cutoff`. I could make 0Hz return a special case where the lowpass is disabled for example

Timple: That's not a bad idea actually. And would also make this PR backward compatible for all controllers not using the D action.

include/path_tracking_pid/details/second_order_lowpass.hpp
```diff
@@ -4,26 +4,41 @@
...
+   * @brief Filter one sample of a signal
+   * @param u signal to be filtered
+   * @param step_size
+   * @return lowpass-filtered signal
+   */
+  double filter(double u, double step_size);
 
-  // Read-only access to the filtered errors FIFO buffer.
-  const FifoArray<double, 3> & filtered_errors() const;
+  void reset();
```
lewie-donckers, line 35: Not all member functions have been documented. Intended? Or did you forget?

include/path_tracking_pid/details/second_order_lowpass.hpp
```diff
@@ -4,26 +4,41 @@
...
-  // Resets both errors and filtered errors FIFO buffers.
-  void reset();
+  /**
+   * @brief Construct a SecondOrderLowpass instance
+   * @param cutoff frequency in Hz, 0 disables the filter
+   * @param damping frequency in Hz
+   */
+  SecondOrderLowpass(double cutoff, double damping);
+
+  void configure(double cutoff, double damping);
```
lewie-donckers, line 25: I see `configure()` is only used together with `reset()`. I can imagine that a `configure()` doesn't make sense without a `reset()`. If so, consider making the configure also do a reset.

include/path_tracking_pid/details/second_order_lowpass.hpp
```diff
@@ -4,26 +4,41 @@
...
 class SecondOrderLowpass
 {
 public:
-  // Pushes the given value to the errors FIFO buffer. A corresponding filtered error value is calculated and pushed
-  // to the filtered errors FIFO buffer.
-  void push(double value);
+  /**
+   * @brief Construct a SecondOrderLowpass instance with NaNs
+   */
+  SecondOrderLowpass();
```
lewie-donckers, line 16: If you change the default member initializer of cutoff and damping, you could default this contructor.

src/details/second_order_lowpass.cpp
```diff
@@ -1,36 +1,48 @@
...
+#include <limits>
 #include <path_tracking_pid/details/second_order_lowpass.hpp>
 
 namespace path_tracking_pid::details
 {
-namespace
-{
-// Used in filter calculations. Default 1.0 corresponds to a cutoff frequency at 1/4 of the sample rate.
-constexpr double cutoff = 1.;
+constexpr auto NaN = std::numeric_limits<float>::quiet_NaN();
```
lewie-donckers, line 7: Why use a float NaN for doubles?

include/path_tracking_pid/details/fifo_array.hpp
```diff
@@ -24,6 +24,9 @@ class FifoArray
   // Read-only access to the element at the given index.
   constexpr const value_type & operator[](std::size_t index) const { return data_[index]; }
 
+  // Read-write access to the element at the given index.
+  value_type & operator[](std::size_t index) { return data_[index]; }
```
lewie-donckers, line 28: As far as I can tell you only use this functionality in `SecondOrderLowpass` because you want a calculation to "look" correct. Basically you push a dummy value and then overwrite it. I don't think that justifies adding this function. Just my 2 cents. If you decide to keep it, add some unit tests for it.

src/details/second_order_lowpass.cpp
```diff
@@ -1,36 +1,48 @@
...
-  return filtered_errors_;
+  // save history
+  u_.push(u);
+  y_.push(u);  // increase index so the math below looks correct
+
+  if (cutoff_ == 0) {
+    return u;
+  }
+
+  auto c = cutoff_;
```
lewie-donckers, line 32: Consider making constants actually `const`. That should be your default. Only remove it, when required.

Also see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

test/unittests/test_second_order_lowpass.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <path_tracking_pid/details/second_order_lowpass.hpp>
+#include <vector>
+
+using path_tracking_pid::details::SecondOrderLowpass;
+
+constexpr double eps = 1e-6;
+
+TEST(SecondOrderLowpass, StepResponse)
+{
+  double dt = 0.1;
```
lewie-donckers, line 13: Consider making constants actually `const`. That should be your default. Only remove it, when required.

Also see https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-immutable

test/unittests/test_second_order_lowpass.cpp
```diff
@@ -0,0 +1,39 @@
...
+  std::vector<double> expected_response = {0.16071,  0.514214, 0.770813, 0.877725, 0.939488,
+                                           0.968659, 0.984211, 0.991911, 0.995898, 0.997907};
+  for (int i = 0; i < static_cast<int>(expected_response.size()); ++i) {
+    SCOPED_TRACE(i);
+
+    auto result = filter.filter(1, dt);
+    EXPECT_NEAR(result, expected_response[i], eps);
+  }
+}
+
```
lewie-donckers, line 28: Consider adding tests with different cutoff and/or damping values. And using different values for dt.

include/path_tracking_pid/details/derivative.hpp
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @brief discrete time derivative filter
+ */
+class Derivative
+{
+public:
+  /**
+   * @brief Construct a Derivative instance
+   */
+  Derivative();
```
lewie-donckers, line 16: Consider removing this. It's not required. The default ctor generated by the compiler does the exact same thing.

include/path_tracking_pid/details/derivative.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * @brief filter one sample of a signal
+   * @param u signal to be filtered
+   * @param step_size
+   * @return derivative of the signal
+   */
+  double filter(double u, double step_size);
+
+  void reset();
```
lewie-donckers, line 26: This does not do anything? Is that intentional? If so, why not just remove it?

If not, consider documenting it.

cfg/Pid.cfg
cesar-lopez-mar: I think Raymon meant the lowpass, for which a 10Hz value sounds good, currently we were using half of the sampling frequency and controllers run by defaults at 20Hz if I am not mistaken. 

We also need only to be careful about what happens when the filter frequency is greater than half the sample frequency? we might through and error or an warning right?

Regarding the damping, I would use 0 as default. In most cases only P action suffices.

include/path_tracking_pid/details/derivative.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <path_tracking_pid/details/fifo_array.hpp>
+
+namespace path_tracking_pid::details
+{
+/**
+ * @brief discrete time derivative filter
+ */
+class Derivative
```
cesar-lopez-mar, line 10: Should we not create also a class for the integral action to make all filter modules generic?.

cesar-lopez-mar: Also a reset to the integral and derivative action can be a good idea after, for instance an emergency stop, the control loop has been opened

src/controller.cpp
```diff
@@ -481,25 +483,17 @@ Controller::UpdateResult Controller::update(
...
-    (controller_state_.error_ang.errors().at<0>() - controller_state_.error_ang.errors().at<1>()) /
-    dt.toSec());
+  auto error_deriv_lat = controller_state_.error_deriv_lat.filter(error_lat_filtered, dt.toSec());
+  auto error_deriv_ang = controller_state_.error_deriv_ang.filter(error_ang_filtered, dt.toSec());
 
   // calculate the control effort
-  const auto proportional_lat =
-    config_.Kp_lat * controller_state_.error_lat.filtered_errors().at<0>();
+  const auto proportional_lat = config_.Kp_lat * error_lat_filtered;
   const auto integral_lat = config_.Ki_lat * controller_state_.error_integral_lat;
```
cesar-lopez-mar, line 491: shall we move this value to just error_integral_lat to use the same structure as the other components?, see previous and next code lines

include/path_tracking_pid/details/second_order_lowpass.hpp
Rayman: Good point, I'll do that

cfg/Pid.cfg
Rayman: Yes that was my point. If we default to 20Hz while the controller runs on 10 Hz, you the filter will has a worse response that if its off.

The lowpass does still work, but it amplifies certain frequencies a bit.

A lowpass damping of `sqrt(2)` should be the default, because that's critically damped. This is a different parameter that the D action.

include/path_tracking_pid/details/second_order_lowpass.hpp
Rayman: These functions have a different function. I'll document this better:
- configure sets the filter parameters. Can be done while driving.
- reset clears the internal buffer to 0. This should be done when starting a new path for example.

I've opened a ticket related to this: https://github.com/nobleo/path_tracking_pid/issues/121

src/details/second_order_lowpass.cpp
Rayman: That's a mistake

include/path_tracking_pid/details/fifo_array.hpp
Rayman: You are correct, technically it is not needed. But it is very helpful if the math corresponds to the code. I'll add some tests for this

include/path_tracking_pid/details/derivative.hpp
Rayman: It's a mistake, it should clear the previous sample

include/path_tracking_pid/details/second_order_lowpass.hpp
Rayman: I've updated the docstrings

include/path_tracking_pid/details/second_order_lowpass.hpp
Rayman: I've done that refactor

src/details/second_order_lowpass.cpp
Rayman: Fixed

test/unittests/test_second_order_lowpass.cpp
Rayman: I've added tests for different damping and cutoff

include/path_tracking_pid/details/derivative.hpp
Rayman: Done

include/path_tracking_pid/details/derivative.hpp
Rayman: Fixed

cfg/Pid.cfg
cesar-lopez-mar: Ahh yes I was confused with the d action of the PID

cesar-lopez-mar: If the controller runs at 10hz then we should not set it above 5hz

include/path_tracking_pid/details/derivative.hpp
Rayman: Good idea. I've created a integral filter with unit tests

src/controller.cpp
Rayman: Good idea. I've created a integral filter with unit tests

src/details/second_order_lowpass.cpp
Rayman: Hmm I don't know. For me adding const here adds clutter. It doesn't add that much because the scope is so small. I can understand from a "purity" type of view this is wrong, but for readability it is good.

Using const in API's I'm fully onboard

test/unittests/test_derivative.cpp
```diff
@@ -0,0 +1,23 @@
...
+  Derivative filter;
+
+  std::vector<double> expected_response = {10, 0, 0};
+  for (int i = 0; i < static_cast<int>(expected_response.size()); ++i) {
+    SCOPED_TRACE(i);
+
+    auto result = filter.filter(1, dt);
+    EXPECT_NEAR(result, expected_response[i], eps);
+  }
+}
```
lewie-donckers, line 23: Consider adding tests for `reset()`.

test/unittests/test_integral.cpp
```diff
@@ -0,0 +1,24 @@
...
+  Integral filter{windup_limit};
+
+  std::vector<double> expected_response = {0.05, 0.15, 0.25, 0.35, 0.45, 0.5, 0.5};
+  for (int i = 0; i < static_cast<int>(expected_response.size()); ++i) {
+    SCOPED_TRACE(i);
+
+    auto result = filter.filter(1, dt);
+    EXPECT_NEAR(result, expected_response[i], eps);
+  }
+}
```
lewie-donckers, line 24: Consider adding tests for the default c'tor, `configure()` and `reset()`. As well as other values for the windup limit.

test/unittests/test_second_order_lowpass.cpp
```diff
@@ -0,0 +1,79 @@
...
+    0.101795, 0.318258, 0.494899, 0.618187, 0.716157,
+    0.786043, 0.84057,  0.880057, 0.91048,  0.932743,
+  };
+  for (int i = 0; i < static_cast<int>(expected_response.size()); ++i) {
+    SCOPED_TRACE(i);
+
+    auto result = filter.filter(1, dt);
+    EXPECT_NEAR(result, expected_response[i], eps);
+  }
+}
```
lewie-donckers, line 79: Consider adding tests for the default c'tor, `configure()` and `reset()`.

test/unittests/test_derivative.cpp
Rayman: Done

test/unittests/test_integral.cpp
Rayman: I've added tests for configure and reset. Windup limit was already tested by the step response (see the last repeating 2 values)

test/unittests/test_second_order_lowpass.cpp
Rayman: Done

lewie-donckers (approved):

cfg/Pid.cfg
Timple: So we should not allow this. In dynamic-reconfigure you can deny (or adjust) a change. This would hold for the controller requency as well as the damping.

Rayman: Right now I've made the default that the lowpass is disabled. If you enable the lowpass, the user can choose a good crossover frequency

Timple: Haha, I should have known that...

src/details/second_order_lowpass.cpp
```diff
@@ -3,34 +3,41 @@
...
+  auto d = damping_;
+  auto T = step_size;
+  auto a = 2 * M_PI * c;
+  auto b = T * a;
+  y_[0] = ((pow(b, 2)) * u_[0] + (2 * pow(b, 2)) * u_[1] + (pow(b, 2)) * u_[2] -
+           (2 * pow(b, 2) - 8) * y_[1] - (pow(b, 2) - 4 * T * a * d + 4) * y_[2]) /
+          (pow(b, 2) + 4 * T * a * d + 4);
```
Rayman, line 34: @MCFurry can you verify that I've made no mistakes in the math for the filter?

This is how I've come to these values:
https://github.com/nobleo/path_tracking_pid/blob/fix/lowpass/doc/second_order_lowpass_tustin.ipynb

src/details/integral.cpp
```diff
@@ -0,0 +1,26 @@
...
+void Integral::configure(double windup_limit) { windup_limit_ = windup_limit; }
+
+double Integral::filter(double u, double step_size)
+{
+  // save history
+  u_.push(u);
+  y_.push(u);  // increase index so the math below looks correct
+
+  auto T = step_size;
+  y_[0] = T / 2 * (u_[0] + u_[1]) + y_[1];
```
rokusottervanger, line 16: This is different math than before, right? Previously, the integral step just added the `new input * T` to the previous output (backward Euler method). Now, you're adding the `average of current and previous inputs * T` to the previous output (trapezoidal method).
 1. Why this change?
 2. Maybe it's good to document the integral method you're using in the header?

Rayman: I've did the same discretezation method (tustin) as the lowpass to prevent confusion. I should definitely add a comment about that.

cfg/Pid.cfg
Rayman: Yes sorry I didn't mention that. The default lowpass crossover frequency is now 0Hz, which means OFF.

Timple (approved):
My comments are addressed. And PR is backwards compatible with P and PI controllers.

src/details/integral.cpp
Rayman: I've added this comment:
> // A continous time integrator was discretized with Tustin's method. For a mathematical
  // explanation, see doc/integral_tustin.ipynb

MCFurry (approved):

rokusottervanger (approved):

src/controller.cpp
```diff
@@ -399,8 +399,9 @@ Controller::UpdateResult Controller::update(
     ROS_WARN("All three gains (Kp, Ki, Kd) should have the same sign for stability.");
   }
 
-  controller_state_.error_lat.push(error.getOrigin().y());
-  controller_state_.error_ang.push(angles::normalize_angle(tf2::getYaw(error.getRotation())));
+  auto error_lat_filtered = controller_state_.error_lat.filter(error.getOrigin().y(), dt.toSec());
+  auto error_ang_filtered = controller_state_.error_ang.filter(
```
Timple, line 403: There is a bracket mismatch here.
Basically this statement now does:
```c++
auto error_ang_filtered = dt.toSec()
```

Rayman: How can this compile, dt is a mandatory argument for filter?

Timple: Sorry, I picked the wrong line :slightly_smiling_face:

Issue #122, opened 2022-03-07
`Controller::distToSegmentSquared()` should be refactored


lewie-donckers:
I think `Controller::distToSegmentSquared()` should be refactored in several ways:

1.  Its name suggests it calculates a distance but the main purpose is to determine a pose: the closest pose on segment. I propose we change the name to reflect this.
1. Beside the pose, it also returns distances but these can easily be calculated by the caller (if required). I propose we remove those distances from the output.
2. It's currently a const member function but it only uses the data member `estimate_pose_angle_enabled_`. I propose we replace that with an input parameter.
3. If the previous point has been done, the function can be extracted (to `calculations.h/cpp` for example) and unit tests can be added.
4. To do this, `distSquared()` from the anonymous namespace in `controller.cpp` would have to be moved too. I propose we do this and add unit tests for that as well.

rokusottervanger:
I approve!

cesar-lopez-mar:
I also agree, it should be more explicit what the function really does.

Issue #127, opened 2022-03-08
`Controller::distToSegmentSquared()` ignores z-component half the time


lewie-donckers:
While working on unit tests for `Controller::distToSegmentSquared()` I noticed that the z-component of the inputs is ignored in some calculations but not in all.

I doubt that this is intended; it looks like a bug but perhaps someone can shed some light on this:

```cpp
  const double l2 = distSquared(pose_v, pose_w); // z used for l2
  if (l2 == 0) {
    return pose_w; // <-- z included in result in this case. not in the other return
  }

  tf2::Transform result;

  const double t = std::clamp(
    ((pose_p.getOrigin().x() - pose_v.getOrigin().x()) *
       (pose_w.getOrigin().x() - pose_v.getOrigin().x()) +
     (pose_p.getOrigin().y() - pose_v.getOrigin().y()) *
       (pose_w.getOrigin().y() - pose_v.getOrigin().y())) /
      l2,
    0.0, 1.0); // <-- z used in l2 but not in rest of calculation of t. which in turn is used for x and y of result.
  result.setOrigin(tf2::Vector3(
    pose_v.getOrigin().x() + t * (pose_w.getOrigin().x() - pose_v.getOrigin().x()),
    pose_v.getOrigin().y() + t * (pose_w.getOrigin().y() - pose_v.getOrigin().y()), 0.0)); // <-- z forced to 0 in result

  const auto yaw = estimate_pose_angle ? atan2(
                                           pose_w.getOrigin().y() - pose_v.getOrigin().y(),
                                           pose_w.getOrigin().x() - pose_v.getOrigin().x())
                                       : tf2::getYaw(pose_v.getRotation());
  tf2::Quaternion pose_quaternion;
  pose_quaternion.setRPY(0.0, 0.0, yaw);
  result.setRotation(pose_quaternion);

  return result;
```

rokusottervanger:
@cesar-lopez-mar Could you shine your light on this?

cesar-lopez-mar:
We definitely should not use z components, unless we explicitly support also commands outside of the plane, which we do not want imo.

PR #130, opened 2022-03-10
Moved distSquared and added tests.

lewie-donckers:
Partial fix for #122 .

Moved `distSquared()` from the anonymous namespace in `controller.cpp` to `calculations.hpp/cpp` and added unit tests.


Timple (approved):
In https://github.com/nobleo/path_tracking_pid/issues/127 it's suggested to drop the Z-component. 
But out of scope for this PR.

Issue #128, opened 2022-03-08
Documentation of `Controller::distToSegmentSquared()` incorrect


lewie-donckers:
While working on unit tests for `Controller::distToSegmentSquared()` I noticed the following.

The documentation suggests that the pose it returns is the point on the segment PV (start, end) that is closest to W (point). It does not do this.

I had the following test case:
```cpp
TEST(PathTrackingPidCalculations, ClosestPointOnSegment_CloseToStart)
{
  const auto start = create_transform(2, 2, 0);
  const auto end = create_transform(4, 4, 0);
  const auto point = create_transform(0, 1, 0);
  const auto ref = start;

  EXPECT_EQ(ref, closestPointOnSegment(start, end, point, false));
}
```

`point` is closest to `start` of any point on `start`-`end` but the following is returned: `[1.759, 2.319, 0]`.

rokusottervanger:
Looks like a nice catch to me.

@cesar-lopez-mar is this expected behavior?

lewie-donckers:
Oops. Might be false alarm. We messed up the documentation and because of that I messed up the order of the arguments in my unit tests. Will fix the documentation and tests and see what happens tomorrow.

rokusottervanger:
Ah, I did the math by hand for this test case. It should be exactly `[1.760, 2.320, 0]`, so that checks out.

lewie-donckers:
It's just a documentation "bug". I fixed it in PR #133 .

PR #133, opened 2022-03-10
Move, refactor and test closestPointOnSegment().

lewie-donckers:
Partial fix for #122 
Fix for #128 (documentation only).

Moved `Controller::closestPointOnSegment()` to `calculations.hpp/cpp` and added unit tests. Found a bug in the documentation (see #128), fixed the documentation and renamed the arguments to make their purpose more clear.

### DIFF
--- a/src/calculations.cpp
+++ b/src/calculations.cpp
@@ -78,7 +78,7 @@ double distSquared(const tf2::Transform & a, const tf2::Transform & b)
   return a.getOrigin().distance2(b.getOrigin());
 }
 
-tf2::Transform closestPointOnSegment(
+tf2::Transform closestPoseOnSegment(
   const tf2::Transform & point, const tf2::Transform & segment_start,
   const tf2::Transform & segment_end, bool estimate_pose_angle)
 {
--- a/src/calculations.hpp
+++ b/src/calculations.hpp
@@ -56,9 +56,9 @@ bool is_in_direction_of_target(
 double distSquared(const tf2::Transform & a, const tf2::Transform & b);
 
 /**
- * @brief Closest point between a line segment and a point
+ * @brief Closest pose between a line segment and a point
  *
- * Calculate the closest point between the line segment bounded by `segment_start` and `segment_end`
+ * Calculate the closest pose between the line segment bounded by `segment_start` and `segment_end`
  * and point `point`.
  *
  * @param[in] point               The point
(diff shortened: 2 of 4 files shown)


src/calculations.hpp
```diff
@@ -55,4 +55,22 @@ bool is_in_direction_of_target(
...
+ * @brief Closest point between a line segment and a point
+ *
+ * Calculate the closest point between the line segment bounded by `segment_start` and `segment_end`
+ * and point `point`.
+ *
+ * @param[in] point               The point
+ * @param[in] segment_start       Start of the line segment
+ * @param[in] segment_end         End of the line segment
+ * @param[in] estimate_pose_angle Indicates if the pose angle should be estimated from the line
+ *                                segment (true) or if the pose angle from segment_start should be
```
cesar-lopez-mar, line 68: I just realized that probably a better idea is to interpolate the angle from start to end (when those angles are given). This will avoid "step" discontinuities of the returned angle in this function between segments (which I saw recently in one of our projects). I will create an issue for this

cesar-lopez-mar (approved):

cesar-lopez-mar (changes requested):
I also realized, closestPoseOnSegment is a better name for this function

src/calculations.cpp
```diff
@@ -76,4 +78,41 @@ double distSquared(const tf2::Transform & a, const tf2::Transform & b)
   return a.getOrigin().distance2(b.getOrigin());
 }
 
+tf2::Transform closestPoseOnSegment(
+  const tf2::Transform & point, const tf2::Transform & segment_start,
+  const tf2::Transform & segment_end, bool estimate_pose_angle)
+{
+  const double l2 = distSquared(segment_start, segment_end);
+  if (l2 == 0) {
```
Timple, line 86: Is it safe to compare a double to an exact value?

lewie-donckers: In this case, I think it is.

We need `l2` not to be equal to 0 so we can divide by it to calculate `t`. If `l2` is nearly 0 but not quite, `t` would be very large but it is clamped between 0 and 1. So, as far as I can tell, the check should not be an issue.

cesar-lopez-mar (approved):
Apart from this feedack (https://github.com/nobleo/path_tracking_pid/pull/133#discussion_r825722520), it looks good to me

Timple (approved):

src/calculations.cpp
rokusottervanger: Also, this is not a change. The implementation is the same as it was before this PR, just moved.

PR #136, opened 2022-03-16
Replaced path pose parameter with return value.

### LINKED ISSUE #123: `Controller::findPositionOnPlan()` should be refactored
`Controller::findPositionOnPlan()` should be refactored:
1. Its name suggests it only finds a position but it also sets member variable `distance_to_goal_`. It is incorrect (or at least odd) to only change the distance but not any other state (current plan index, etc). I propose we make this function const and do not set any data members. It should just do as the name suggests and find something. If required this value can be made part of the return value. (But perhaps it might be easy to calculate by the caller.)
2. Parameter `path_pose_idx` is an output parameter. I propose we change it to be part of the return value.
3. Parameter `controller_state` contains way too much state. Only the following members are used:
    - `last_visited_pose_index` is used as an output parameter. I propose we change it to be part of the return value.
    - `current_global_plan_index` is used as an in-/output parameter. I propose we split this into a pure input parameter and make the output part of the return value.

lewie-donckers:
Next PR to fix #123 .

Refactored `Controller::findPositionOnPlan()`: replaced `path_pose_idx` output parameter with (part of) the return value.

### DIFF
--- a/include/path_tracking_pid/controller.hpp
+++ b/include/path_tracking_pid/controller.hpp
@@ -96,22 +96,21 @@ public:
     const geometry_msgs::Twist & steering_odom_twist,
     const std::vector<tf2::Transform> & global_plan);
 
+  /** Result of findPositionOnPlan(). */
+  struct FindPositionOnPlanResult
+  {
+    tf2::Transform position;
+    std::size_t path_pose_idx = 0;
+  };
+
   /**
    * Find position on plan by looking at the surroundings of last known pose.
    * @param current Where is the robot now?
    * @param controller_state  The current state of the controller that gets updated by this function
-   * @return tf of found position on plan
-   * @return index of current path-pose if requested
+   * @return Found position on plan and the index of current path-pose
    */
-  tf2::Transform findPositionOnPlan(
-    const tf2::Transform & current_tf, ControllerState & controller_state, size_t & path_pose_idx);
-  // Overloaded function definition for users that don't require the segment index
-  tf2::Transform findPositionOnPlan(
-    const tf2::Transform & current_tf, ControllerState & controller_state)
-  {
-    size_t path_pose_idx;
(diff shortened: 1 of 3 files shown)


Rayman (approved):

include/path_tracking_pid/controller.hpp
```diff
@@ -96,22 +96,21 @@ class Controller : private boost::noncopyable
     const geometry_msgs::Twist & steering_odom_twist,
     const std::vector<tf2::Transform> & global_plan);
 
+  /** Result of findPositionOnPlan(). */
+  struct FindPositionOnPlanResult
+  {
+    tf2::Transform position;
```
rokusottervanger, line 102: I'd rather see this member named `pose` instead of `position`, as it represents not only a position, but also the orientation of the projection onto the path (together, this is called the "pose"). Maybe this also implies that the function should be named `FindPoseOnPlan` or something. Does that make sense, @cesar-lopez-mar ?

rokusottervanger (changes requested):
Other than the single comment, I approve!

include/path_tracking_pid/controller.hpp
cesar-lopez-mar: Indeed, also the data type represents a pose indirectly, we use transforms because we use functions of tf2 objects to make operations on them. When track_base_link is false we do not use the orientation but the same data type is used.

lewie-donckers: Fixed in #144 to avoid merge conflicts in the PR train. Since the name was not changed yet, it could be considered just another refactor step.

rokusottervanger: Sounds fair enough

rokusottervanger (approved):

cesar-lopez-mar (approved):

Issue #145, opened 2022-03-17
Controller's end phase distance is calculated wrong for backwards motion


PaulVerhoeckx:
`d_end_phase` ([here](https://github.com/nobleo/path_tracking_pid/blob/dbe631d210f389466cc3069651588aff1ce39818/src/controller.cpp#L448-L460)) can result in negative values if `target_x_vel` < 0

cesar-lopez-mar:
I think the issue is not whether d_end_phase is negative but how its value is used. Negative d_end_phase just means that the robot will end it's current phase at a position behind the current position

PR #150, opened 2022-03-22
fix: Use absolute target velocity to calculate end phase distance

PaulVerhoeckx:
Solves #145

### DIFF
--- a/src/controller.cpp
+++ b/src/controller.cpp
@@ -448,16 +448,12 @@ Controller::UpdateResult Controller::update(
   if (
     (current_target_x_vel_ > 0.0 && current_x_vel > target_end_x_vel) ||
     (current_target_x_vel_ < 0.0 && current_x_vel < target_end_x_vel)) {
-    t_end_phase_current = (target_end_x_vel - current_x_vel) / (-config_.target_x_decc);
-    d_end_phase = current_x_vel * t_end_phase_current -
-                  0.5 * (config_.target_x_decc) * t_end_phase_current * t_end_phase_current +
-                  fabs(target_x_vel) * 2.0 * dt.toSec();
+    t_end_phase_current = fabs((target_end_x_vel - current_x_vel) / config_.target_x_decc);
   } else {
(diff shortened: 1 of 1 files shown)


cesar-lopez-mar:
I think we should check how d_end_phase is used instead of changing is calculation. Negative d_end_phase is allowed, indicating the robot will end the current section at a distance behind the current position

PaulVerhoeckx:
I'm pretty sure the implementation is wrong. `d_end_phase` is set equal to the current theoretical stopping distance plus some factor which should compensate for delays. The sign of this theoretical stopping distance does not depend on the sign of `current_x_vel` while the compensation part does.

For example, take:
`target_end_x_vel = 0`
`current_x_vel = target_x_vel = 1`
`target_x_decc = 1`
`dt=1/20`

This yields:
```
t_end_phase_current = (target_end_x_vel - current_x_vel) / (-target_x_decc) = 1
```
```
d_end_phase = current_x_vel * t_end_phase_current -
                  0.5 * (target_x_decc) * t_end_phase_current * t_end_phase_current +
                  target_x_vel * 2.0 * dt = 1 - 0.5 + 0.1 = 0.6
```

While for when driving backwards, with:
`current_x_vel = target_x_vel = -1`

This yields:
```
t_end_phase_current = (target_end_x_vel - current_x_vel) / (-target_x_decc) = -1
```
```
d_end_phase = current_x_vel * t_end_phase_current -
                  0.5 * (target_x_decc) * t_end_phase_current * t_end_phase_current +
                  target_x_vel * 2.0 * dt = 1 - 0.5 - 0.1 = 0.4
```

cesar-lopez-mar:
> t_end_phase_current = (target_end_x_vel - current_x_vel) / (-target_x_decc) = -1
Time should not be negative. The bug is in this equation. During braking when going backwards we are actually accelerating! so the minus sign we added here should depend on the sign of the target velocity

PaulVerhoeckx:
I proposed a different fix to the problem in a new commit

